PR #1, opened 2018-07-25
User creation capability

felipemontoya:
## Description
This PR adds the capacity to create users on a real & running edxapp.

Direct import from edxapp are done on selectable backends that should be tested against the specific branch of edx-platform that they target.
To select a backend the settings module from django conf and the plugin is used.

## Reviewers
- [x] @Ivanca 

## Pending
- [x] Add tests for the public api function
- [ ] Add tests for the backends

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,16 +46,11 @@ jobs:
             - ./venv
           key: v1-dependencies-{{ checksum "requirements.txt" }}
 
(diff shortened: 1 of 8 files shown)


felipemontoya:
@Ivanca This is ready to merge on my side. It passes the unit test for the internal module and creates a very rough backend that works as a POC.

PR #4, opened 2018-07-26
Remove "plugins" and "testsite" folders

Ivanca:


felipemontoya:
I'm going to merge this even with failing tests, because the directories need to disappear. Tests will be fixed on PR #5

PR #7, opened 2018-07-26
Add coverage.py to the eox-core project (incl. circleci config)

Ivanca:


felipemontoya:
@Ivanca I did some changes on the circle config in PR #1. Since that has the first tests, we can merget that and then rebase this and make sure that coverage works with the DJANGO_SETTINGS_MODULE definition.

PR #9, opened 2018-07-27
Adding git sha to the output of eox-info

felipemontoya:
This PR adds extra information to the info view

The git sha is there to test the integrity of the version number or to provide extra granularity in case the installed version does not match exactly with a release


felipemontoya:
@Ivanca the docstrings are in now

PR #12, opened 2018-07-30
Correcting git sha

felipemontoya:
This PR fixes the call to the git cmd line so that is actually returns the ID of the commit in the eox-core package instead of that of edx-platform


felipemontoya:
@Ivanca can you give me a review here, so that we can start closing branches please?

Ivanca:
ᶘ ᵒᴥᵒᶅ seal of approval

PR #15, opened 2018-08-08
Enrollment api

Ivanca:


eox_core/api/v1/serializers.py
```diff
@@ -46,3 +46,35 @@ class EdxappUserQuerySerializer(EdxappUserSerializer):
...
+    VERIFIED = 'verified'
+    AUDIT = 'audit'
+    NO_ID_PROFESSIONAL_MODE = 'no-id-professional'
+    CREDIT_MODE = 'credit'
+    ALL_MODES = [AUDIT, CREDIT_MODE, HONOR, NO_ID_PROFESSIONAL_MODE, PROFESSIONAL, VERIFIED, ]
+
+    user = serializers.CharField(max_length=30)
+    created = serializers.DateTimeField(allow_null=True)
+    is_active = serializers.BooleanField(default=True)
+    mode = serializers.CharField(default=AUDIT, max_length=100)
```
felipemontoya: @Ivanca does it make sense to use `choices` here?

eox_core/api/v1/views.py
```diff
@@ -14,8 +14,9 @@
 
 from rest_framework.authentication import BasicAuthentication
 from rest_framework.permissions import AllowAny
-from eox_core.api.v1.serializers import EdxappUserQuerySerializer, EdxappUserSerializer
+from eox_core.api.v1.serializers import EdxappUserQuerySerializer, EdxappUserSerializer, EdxappCourseEnrollmentSerializer
```
felipemontoya: This is just going to keep growing, how about moving to the syntax:
```
from eox_core.api.v1.serializers import (
    ....
)
```

eox_core/edxapp_wrapper/backends/enrollment_h_v1.py
```diff
@@ -0,0 +1,9 @@
+
```
felipemontoya: do we need to import print_function from futures here?

eox_core/edxapp_wrapper/enrollments.py
```diff
@@ -0,0 +1,26 @@
...
+def create_enrollment(*args, **kwargs):
+    """ Creates the edxapp user """
+
+    backend_function = settings.EOX_CORE_ENROLLMENT_BACKEND
+    backend = import_module(backend_function)
+
+    return backend.create_enrollment(*args, **kwargs)
+
+
+def check_edxapp_account_conflicts(*args, **kwargs):
```
felipemontoya: this function check_edxapp_account_conflicts wont be necessary here

eox_core/edxapp_wrapper/backends/enrollment_h_v1.py
Ivanca: print is not used anymore on the last commit (a08e8c4)

eox_core/api/v1/serializers.py
```diff
@@ -46,3 +48,43 @@ class EdxappUserQuerySerializer(EdxappUserSerializer):
...
+class EdxappCourseEnrollmentSerializer(serializers.Serializer):
+    """Serializes CourseEnrollment
+
+    Aggregates all data from the Course Enrollment table, and pulls in the serialization for
+    the Course Descriptor and course modes, to give a complete representation of course enrollment.
+
+    """
+
+    user = serializers.CharField(max_length=30, default="")
+    #created = serializers.DateTimeField(default=datetime.now)
```
felipemontoya: You can safely delete this

eox_core/api/v1/serializers.py
```diff
@@ -46,3 +48,43 @@ class EdxappUserQuerySerializer(EdxappUserSerializer):
...
+
+class EdxappCourseEnrollmentQuerySerializer(EdxappCourseEnrollmentSerializer):
+    """
+    Handles the serialization of the context data required to create an enrollemnt
+    on different backends
+    """
+    username = serializers.CharField(max_length=30)
+    force_registration = serializers.BooleanField(default=False)
+    course_id = serializers.CharField(max_length=255)
+    enrollment_attributes = EdxappEnrollmentAttributeSerializer(many=True)
```
felipemontoya: @Ivanca do you have an example of what kind of enrollment_attributes this will hold?

eox_core/edxapp_wrapper/backends/enrollment_h_v1.py
```diff
@@ -0,0 +1,54 @@
...
+            api.set_enrollment_attributes(user_id, course_id, enrollment_attributes)
+    except (NonExistentCourseError, EnrollmentClosedError, CourseFullError, AlreadyEnrolledError) as e:
+        enrollment = None
+        errors.append(repr(e))
+    return enrollment, errors
+
+
+def check_edxapp_enrollment_is_valid(*args, **kwargs):
+    errors = []
+    is_active = kwargs.get("force_registration", False)
```
felipemontoya: likely an error is_active <=> force_registration

felipemontoya: or just duplicated and we can leave the one on L#45

eox_core/edxapp_wrapper/backends/enrollment_h_v1.py
```diff
@@ -0,0 +1,54 @@
...
+
+def create_enrollment(*args, **kwargs):
+
+    errors = []
+    user_id = kwargs.get('username')
+    course_id = kwargs.get('course_id')
+    mode = kwargs.get('mode')
+    is_active = kwargs.get('is_active', True)
+    enrollment_attributes = kwargs.get('enrollment_attributes', None)
+    validation_errors = check_edxapp_enrollment_is_valid(*args, **kwargs)
```
felipemontoya: What do you think of moving this into the serializer validation method? This way if the `check_edxapp_enrollment_is_valid` fails we inmediately return from the serializer in the class EdxappEnrollment#L60

Ivanca: Serializer validation method already calls this method too. This way of doing in both part is what /v1/users/ does.

eox_core/api/v1/serializers.py
Ivanca: https://github.com/edx/edx-platform/blob/master/common/djangoapps/enrollment/views.py#L430-L435

eox_core/api/v1/serializers.py
Ivanca: done

eox_core/edxapp_wrapper/backends/enrollment_h_v1.py
Ivanca: done (removed duplicate)

PR #17, opened 2018-08-09
Allow lookup by email on the eox-core/enrollment API

Ivanca:


eox_core/edxapp_wrapper/backends/enrollment_h_v1.py
```diff
@@ -30,6 +32,13 @@ def create_enrollment(*args, **kwargs):
     validation_errors = check_edxapp_enrollment_is_valid(*args, **kwargs)
     if validation_errors:
         return None, [", ".join(validation_errors)]
+    if email != "":
+        match = get_user_model().objects.filter(email=email).first()
+        if match is None:
+            raise APIException('no user found with that email')
```
felipemontoya: Capital `N`

eox_core/edxapp_wrapper/backends/enrollment_h_v1.py
```diff
@@ -30,6 +32,13 @@ def create_enrollment(*args, **kwargs):
     validation_errors = check_edxapp_enrollment_is_valid(*args, **kwargs)
     if validation_errors:
         return None, [", ".join(validation_errors)]
+    if email != "":
```
felipemontoya: This essentially gives higher precedence to the email. Do we want this? what if the api call includes both username and email, what if the username and the email point to a different user object?

eox_core/api/v1/serializers.py
```diff
@@ -83,7 +81,8 @@ class EdxappCourseEnrollmentQuerySerializer(EdxappCourseEnrollmentSerializer):
     Handles the serialization of the context data required to create an enrollemnt
     on different backends
     """
-    username = serializers.CharField(max_length=30)
+    username = serializers.CharField(max_length=30, default="")
```
felipemontoya: Can we default to `None` instead of an empty string? This way we are not comparing to emtpy strings all over the code

eox_core/edxapp_wrapper/backends/enrollment_h_v1.py
Ivanca: Added an error when both are send

eox_core/api/v1/serializers.py
Ivanca: Changed to None but FWIW empty strings evaluate to false in python (just like None does)

eox_core/edxapp_wrapper/backends/enrollment_h_v1.py
Ivanca: done

Ivanca:
Another review pending after changes done on 11aed0f @felipemontoya

eox_core/edxapp_wrapper/backends/enrollment_h_v1.py
```diff
@@ -17,11 +17,13 @@
...
 from course_modes.models import CourseMode
 from rest_framework.exceptions import APIException
+from django.contrib.auth import get_user_model
 
 
 def create_enrollment(*args, **kwargs):
 
     errors = []
+    email = kwargs.get("email")
     user_id = kwargs.get('username')
```
felipemontoya: final nit: you assign username to the user_id. That is a bit confusing. Why not user `username = kwargs.get('username')`

eox_core/edxapp_wrapper/backends/enrollment_h_v1.py
```diff
@@ -46,11 +55,15 @@ def check_edxapp_enrollment_is_valid(*args, **kwargs):
     course_id = kwargs.get("course_id")
     force_registration = kwargs.get('force_registration', False)
     mode = kwargs.get("mode")
+    if not kwargs.get("email") and not kwargs.get("username"):
+        return ['Email or username needed']
+    if kwargs.get("email") and kwargs.get("username"):
+        return ['Email or username but not both']
```
felipemontoya: A bit more context in the message would be better: 'You have to provide an email or username but not both'

eox_core/edxapp_wrapper/backends/enrollment_h_v1.py
Ivanca: Changed. But the logic was because the functions we are calling (create_course_enrollment) always call it "user_id" in their definitions

eox_core/edxapp_wrapper/backends/enrollment_h_v1.py
Ivanca: Ok done

PR #18, opened 2018-08-13
Allow Json array for /enrollment/ api, allow update if force_registration is true

Ivanca:


eox_core/edxapp_wrapper/backends/enrollment_h_v1.py
```diff
@@ -38,13 +39,16 @@ def create_enrollment(*args, **kwargs):
...
         enrollment = api._data_api().create_course_enrollment(username, course_id, mode, is_active)
         if enrollment_attributes is not None:
             api.set_enrollment_attributes(username, course_id, enrollment_attributes)
-    except (UserNotFoundError, InvalidEnrollmentAttribute, CourseNotFoundError, CourseEnrollmentFullError, CourseEnrollmentClosedError, CourseEnrollmentExistsError) as e:
-        enrollment = None
+    except CourseEnrollmentExistsError as e:
+        if force_registration:
+            enrollment = api._data_api().update_course_enrollment(username, course_id, mode, is_active)
+        else:
+            raise APIException(repr(e) + ", use force_registration=1 if updating is what you need")
```
felipemontoya: "use force_registration to update the existing enrollment"

Ivanca: Done

Ivanca:
Separated the small change into another PR #19 (icc/update-enrollment-if-force_registration)

eox_core/api/v1/views.py
```diff
@@ -56,16 +55,27 @@ def post(self, request, *args, **kwargs):
...
+        for one in data:
+            enrollment, msg = create_enrollment(**one)
+            serialized_enrollment = EdxappCourseEnrollmentSerializer(enrollment)
+            response_data = serialized_enrollment.data
+            if msg:
+                response_data["messages"] = msg
+            multiple_responses.append(response_data)
+
+        if many:
+            return Response(multiple_responses)
```
felipemontoya: Just a reminder of the discussion we had per other channel to make this a bit cleaner

Ivanca: @felipemontoya Should I make the change that we discussed here (the decorator) or is that for a later refactor?

felipemontoya: If you can manage to do that in a few hours tops, you can add it here. If not, then merge now and we will use this code as it is for a few years ;)

felipemontoya:
@Ivanca one thing that we do need to address now. The name of the flag.

eox_core/api/v1/serializers.py
```diff
@@ -85,4 +85,4 @@ class EdxappCourseEnrollmentQuerySerializer(EdxappCourseEnrollmentSerializer):
     email = serializers.CharField(max_length=255, default=None)
     force_registration = serializers.BooleanField(default=False)
```
felipemontoya: Can we make it "force" instead of "force_registration"? If we do, does the WP plugin already uses `force_registration` and we require a change there as well?

Ivanca: No change needed on the WP plugin

felipemontoya: cool, then I say, lets do it and merge this

PR #21, opened 2018-08-30
Fixing pylint complains

Ivanca:
And adding pylint to "make test" so it executes automatically.


Ivanca:
@felipemontoya any change required here?

PR #23, opened 2018-09-14
wip: create_program_enrollment

Ivanca:


eox_core/edxapp_wrapper/backends/edxfuture.py
```diff
@@ -0,0 +1,33 @@
+"""
```
felipemontoya: a este archivo puedes llamarlo porfa edxfuture_i_v1 ?

eox_core/edxapp_wrapper/backends/enrollment_h_v1.py
```diff
@@ -8,21 +8,27 @@
...
+from django.contrib.sites.models import Site
+from django.db import models
 from enrollment import api
 from enrollment.errors import CourseModeNotFoundError
 from enrollment.errors import CourseEnrollmentExistsError
 from course_modes.models import CourseMode
 from opaque_keys.edx.keys import CourseKey
 from student.models import CourseEnrollment
+from edxfuture import get_program
+import pprint
```
felipemontoya: Just marking so we dont forget to remove the pprint before merging

eox_core/edxapp_wrapper/backends/enrollment_h_v1.py
```diff
@@ -8,21 +8,27 @@
...
 
 LOG = logging.getLogger(__name__)
 
-
 def create_enrollment(*args, **kwargs):
     """
     backend function to create enrollment
     """
     errors = []
+    program_uuid = kwargs.get('bundle_id') or kwargs.get('program_id')
```
felipemontoya: I would leave it as bundle_id only.

eox_core/edxapp_wrapper/backends/enrollment_h_v1.py
```diff
@@ -8,21 +8,27 @@
...
+    if program_uuid:
```
felipemontoya: this if condition here is weird. It is the classic example of a function that was extended for something not meant for it.
(which is what we are doing anyways)

What if we have `create_enrollment` be the general function, and according to the existance of `bundle_id` or `course_id` we call either `enroll_on_course` or `enroll_on_program`

felipemontoya: The result would naturally be the same for the caller of the API, but reading this code would be a lot cleaner

eox_core/edxapp_wrapper/backends/enrollment_h_v1.py
```diff
@@ -54,6 +60,20 @@ def create_enrollment(*args, **kwargs):
...
+def create_program_enrollment(program_uuid, *arg, **kwargs):
+    """
+    backend function to create enrollment
+    """
+    try:
+        site = models.ForeignKey(Site)
+        data = site.siteconfiguration.get_program(program_uuid)
+        for x in data['courses']:
+            print(x)
+            pprint(vars(x))
```
felipemontoya: I suppose here we are going to do the enrollments one by one. Maybe this should also call the `enroll_on_course` function at some point

eox_core/edxapp_wrapper/backends/edxfuture.py
Ivanca: Ok

eox_core/edxapp_wrapper/backends/enrollment_h_v1.py
Ivanca: Ok

eox_core/edxapp_wrapper/backends/enrollment_h_v1.py
Ivanca: Done

eox_core/edxapp_wrapper/backends/enrollment_h_v1.py
Ivanca: Sure

eox_core/edxapp_wrapper/backends/enrollment_h_v1.py
Ivanca: Yeah that was the idea

eox_core/edxapp_wrapper/backends/enrollment_h_v1.py
```diff
@@ -22,10 +24,19 @@ def create_enrollment(*args, **kwargs):
     """
     backend function to create enrollment
     """
+    program_uuid = kwargs.get('bundle_id')
+    course_id = kwargs.get('course_id')
+
+    if program_uuid:
+        return enroll_on_program(program_uuid, *args, **kwargs)
+    else:
```
felipemontoya: What if the bundle_id or course_id don't exist?
On the other hand this would be safer being two separate conditions, the existence of program_uuid and the existence of course_id

eox_core/edxapp_wrapper/backends/edxfuture.py
```diff
@@ -0,0 +1,38 @@
...
+Functions copied from a version higer than hawtorn (for backwards compatibility with i)
```
felipemontoya: typo on hawthorn

PR #25, opened 2018-09-18
Add logging on enroll API

Ivanca:
-

### DIFF
--- a/eox_core/api/v1/views.py
+++ b/eox_core/api/v1/views.py
@@ -3,6 +3,7 @@ API v1 views.
 """
 
 from __future__ import absolute_import, unicode_literals
+import logging
 
 from rest_framework.views import APIView
 from rest_framework.response import Response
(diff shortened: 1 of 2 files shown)


eox_core/api/v1/views.py
```diff
@@ -73,10 +77,35 @@ def post(self, request, *args, **kwargs):
...
                 multiple_responses.append(response_data)
-
+        for item in data:
+            self.log(item)
         response = multiple_responses if many else multiple_responses[0]
         return Response(response)
 
+    def handle_exception(self, exc):
+        """
+        Handle execption: log it
```
felipemontoya: typo

eox_core/edxapp_wrapper/backends/enrollment_h_v1.py
```diff
@@ -76,12 +79,14 @@ def enroll_on_program(program_uuid, *arg, **kwargs):
...
     results = []
     errors = []
+    LOG.info('Enrolling on program: %s', program_uuid)
     try:
         data = get_program(program_uuid)
     except Exception as err:  # pylint: disable=broad-except
         raise APIException(repr(err))
     for course in data['courses']:
         if course['course_runs']:
+            LOG.info('Enrolling on course: %s', course['key'])
```
felipemontoya: This message should be about the process we are in. So "enrolling on course_run: xxxx" and it should bo after we have run `get_preferred_course_run`

PR #33, opened 2018-10-19
Making sure that if no courses are available we fail verbosely

felipemontoya:
This PR is the result of debugging a particular behavior on production for a given program


felipemontoya:
thanks @Ivanca

PR #35, opened 2018-11-23
"direnv allow" missing on README.md

Ivanca:
"direnv allow" missing on README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,7 @@ make dev.checkout
 make dev.clone
 make dev.provision
 echo export OPENEDX_RELEASE=hawthorn.master > .envrc
+direnv allow
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -40,6 +40,7 @@ make dev.checkout
 make dev.clone
 make dev.provision
 echo export OPENEDX_RELEASE=hawthorn.master > .envrc
+direnv allow
```
felipemontoya: This migh work for you personally, but not everyone uses direnv, so I think it would create more clutter. Ironically

felipemontoya:
In my opinion we should not merge this

felipemontoya:
This has not been needed for a while, so I'm closing the PR

PR #36, opened 2018-11-29
Retrieve user-info GET end-point

Ivanca:
By email or/and username, example of how to use it at:
```bash
curl --header "Authorization: Bearer f04829ae958e77dd8c7015695a82062890ec9fe4" -H "Accept: application/json" \
        "http://localhost:18000/eox-core/api/v1/user/?username=example_user" 
```


eox_core/edxapp_wrapper/backends/users_h_v1.py
```diff
@@ -6,14 +6,37 @@
...
+from openedx.core.djangoapps.user_api.accounts.serializers import UserReadOnlySerializer  # pylint: disable=import-error
 from student.forms import AccountCreationForm  # pylint: disable=import-error
 from student.helpers import do_create_account, create_or_set_user_attribute_created_on_site  # pylint: disable=import-error
 from student.models import create_comments_service_user  # pylint: disable=import-error
 
 LOG = logging.getLogger(__name__)
+User = get_user_model()  # pylint: disable=invalid-name
+
+
+def get_edxapp_user(**kwargs):
```
felipemontoya: This function is a cornerstone of many features to come. Since it is very important I would like to make sure this function is aware of the tenancy concept (microsite, site_configurations, or others) and that it only retrieves users if the function is being called at the right tenant.

Ivanca:
Some further improvements done @felipemontoya

eox_core/edxapp_wrapper/backends/users_h_v1.py
Ivanca: Functionality already added.

Makefile
```diff
@@ -25,7 +25,7 @@ clean: ## remove generated byte code, coverage reports, and build artifacts
 	rm -fr dist/
 	rm -fr *.egg-info
 
-test: clean
+test: clean ## Run tests
 	pylint eox_core
-	coverage run --source="." -m pytest
+	coverage run --source="." --omit="*/backends*" -m pytest
```
felipemontoya: Why are you doing this here instead of using the setup.cfg which already has a coverage section?

PR #39, opened 2019-01-22
Changing API permissions

diegomillan:
This PR adds a custom permission to decide if an API request is processed or not 

@felipemontoya 
@Squirrel18 
@jfavellar90


diegomillan:
Just applied your feedback @felipemontoya

PR #45, opened 2019-02-07
LLPA-20 - Course management - Course team add/remove.

Squirrel18:
## Description:

This PR adds most of the React logic that will be used in the course management view.

Also adds the first tab called: Course team management, which handles adding/removing the course team members (staff, instrcutor), for all courses in one organization.

## How to test:

 - Install eox-core, checkout to this branch.
 - Go to eox-core installed path and run: npm run build.
 - Go to <path-to-studio>/admin
 - Add and activate a new Waffle switch called: course_management_automation_module
 - Login into Studio
 - Go to <path-to-studio>eox-core/management/courses

## Future work:

We need to setup Bootstrap correctly, in order to keep the base.html file styles as they come and add the required styles only to the course management view.

## Reviewers:

 - [ ] @amalbas 
 - [ ] @diegomillan 
 - [ ] @felipemontoya


eox_core/cms/static/js/index.jsx
```diff
@@ -1,10 +1,38 @@
...
-    return (
-      <h1>Hello React js from eox-core.</h1>
+function CourseManagement(props) {
+  return (
+    <Tabs labels={["Course team management", "Panel 2", "Panel 3"]}>
+      <div>
+        <CourseTeamManagement orgList={props.teamManagement.list_org}/>
+      </div>
+      <div>
+        Hello I am the second panel
```
diegomillan: Let's change the dummy text to "Panel X"

eox_core/cms/views.py
```diff
@@ -1,19 +1,154 @@
...
+    if not org:
+        data = {
+            'message': 'Incorrect organization name.',
+            'status': 'Failed'
+        }
+        return JsonResponse(data, status=status.HTTP_400_BAD_REQUEST)
+    elif not active_courses:
+        data = {
+            'message':
+                'The organization {} does not exists, or has not active courses.'.format(org),
```
diegomillan: I think `or does not have active courses` sounds better

eox_core/cms/views.py
```diff
@@ -1,19 +1,154 @@
...
+def get_json_from_body_request(request):
+    """
+    Function to convert request.body into a json format.
+    """
+    json_content = {}
+    if "application/json" in request.META.get('CONTENT_TYPE', '') and request.body:
+        try:
+            json_content = json.loads(request.body)
+        except ValueError:
+            return json_content
```
diegomillan: Why return  `json_content` even if there is a `ValueError`?

eox_core/edxapp_wrapper/backends/users_h_v1.py
```diff
@@ -4,19 +4,28 @@
...
-from django.db import transaction
-from django.contrib.auth import get_user_model
+from contentstore.views.user import _course_team_user
 from django.conf import settings
-from rest_framework.exceptions import APIException
-from openedx.core.djangoapps.lang_pref import LANGUAGE_KEY  # pylint: disable=import-error
-from openedx.core.djangoapps.user_api.accounts.api import check_account_exists  # pylint: disable=import-error
+from django.contrib.auth import get_user_model
+from django.db import transaction
+from openedx.core.djangoapps.lang_pref import \
```
diegomillan: I would recommend to import in two lines like this:
```python
from openedx.core.djangoapps.lang_pref import (
    Mymodule
)
```

Applies for all of them

eox_core/edxapp_wrapper/backends/users_h_v1.py
```diff
@@ -4,19 +4,28 @@
...
+    check_account_exists  # pylint: disable=import-error
+from openedx.core.djangoapps.user_api.accounts.serializers import \
+    UserReadOnlySerializer  # pylint: disable=import-error
 from openedx.core.djangoapps.user_api.preferences import api as preferences_api  # pylint: disable=import-error
-from openedx.core.djangoapps.user_api.accounts.serializers import UserReadOnlySerializer  # pylint: disable=import-error
+from rest_framework.exceptions import APIException
 from student.forms import AccountCreationForm  # pylint: disable=import-error
-from student.helpers import do_create_account, create_or_set_user_attribute_created_on_site  # pylint: disable=import-error
-from student.models import create_comments_service_user, UserAttribute, UserSignupSource, CourseEnrollment  # pylint: disable=import-error
+from student.helpers import \
```
diegomillan: Same as above

diegomillan:
This is looking good so far, butI think @amalbas feedback over react components and code structure would be valuable here.

eox_core/cms/views.py
Squirrel18: Because returns an empty dict. @diegomillan

diegomillan: ok

PR #47, opened 2019-02-20
LLPA-20 - Course management - Course settings.

Squirrel18:
## Description:

This PR adds a second tab called: Course settings, which handles the changes to the advanced settings and Schedule & Details settings for all course that match with the course string regex.

## How to test:

* Install eox-core, checkout to this branch.
* Go to eox-core installed path and run: npm run build.
* Go to /admin
* Add and activate a new Waffle switch called: course_management_automation_module
* Login into Studio
* Go to eox-core/management/courses
* Search courses with the exactly course key or a wildcard * character. e.g. org*_2018
* Select the desired settings, fill the new value and push the button called: Apply changes.

## Screenshot:

Course settings tab:

image

Course team management:

image

## Notes:

* This PR adds the fixed styles to the first tab too.
* If any styles looks broken or could be improved, please let me know before squashing.

## Reviewers:

 - [ ] @diegomillan 
 - [ ] @amalbas 
 - [ ] @felipemontoya


eox_core/cms/views.py
```diff
@@ -31,30 +35,36 @@ def management_view(request):
     """
     if 'text/html' in request.META.get('HTTP_ACCEPT', '') and request.method == 'GET':
```
diegomillan: since we already have a `@require_http_methods(["GET"])`, we could remove `request.method == 'GET'`

Squirrel18:
@diegomillan Feedback addressed.

eox_core/cms/templates/management.html
```diff
@@ -1,23 +1,29 @@
...
 %>
 <%block name="title">${"Course management"}</%block>
 <%block name="header_extras">
 
 </%block>
 
 <%block name="content">
-    <link type='text/css' rel='stylesheet' href='${STATIC_URL}css/vendor/bootstrap.min.css'>
-    <div class="container" style="font-size: 2em;">
+    <link type='text/css' rel='stylesheet' href='${STATIC_URL}bundle/css/course-management.bundle.css'>
```
felipemontoya: normally on the Open edX platform the css static files are loaded using:
`${static.url('xxx/yyy/zzz/file.min.css')}`

Why are we not using that standard here?

eox_core/cms/urls.py
```diff
@@ -1,11 +1,17 @@
 """ urls.py """
 
 from django.conf.urls import url, include
-from eox_core.cms.views import management_view, course_team, get_all_orgs
+from eox_core.cms.views import (
+    management_view,
+    course_team,
+    get_all_orgs,
+    get_courses_by_course_regex
```
felipemontoya: trailing comma.

Are we even running this code through a pycodestyle linter?

eox_core/cms/views.py
```diff
@@ -1,19 +1,24 @@
...
 from django.views.decorators.http import require_http_methods
 from opaque_keys.edx.keys import CourseKey  # pylint: disable=import-error
 from rest_framework import status
-from rest_framework.response import Response
 
 from eox_core.edxapp_wrapper.courses import (get_courses_accessible_to_user,
-                                             get_process_courses_list)
+                                             get_process_courses_list,
+                                             get_course_details_fields,
+                                             get_first_course_key,)
```
felipemontoya: this for example is a place where the trailing comma makes no sense

eox_core/cms/views.py
```diff
@@ -29,32 +34,40 @@ def management_view(request):
     """
     Renders the course management view.
     """
-    if 'text/html' in request.META.get('HTTP_ACCEPT', '') and request.method == 'GET':
+    if 'text/html' in request.META.get('HTTP_ACCEPT', ''):
         org_list = get_all_orgs()
+        details_fields = get_course_details_fields()
+        course_key = get_first_course_key()
+        request_timeout_value = getattr(settings, 'EOX_CORE_COURSE_MANAGEMENT_REQUEST_TIMEOUT', 500)
```
felipemontoya: Since you are already setting a default on the `common.py` module I think this default is not necessary. Also if it was, you are setting a different value. The best option is to define it only once (in the common.py).

eox_core/cms/views.py
```diff
@@ -29,32 +34,40 @@ def management_view(request):
...
-        <studio-url>/eox-core/management/course-team-management
+        POST, DELETE <studio-url>/eox-core/management/course-team-management
 
     **Request Values**
 
         * org: Organization short name.
         * user: user email.
         * role: user role (staff, instructor, '')
 
+        Note: If the role value is empty must be send it as a DELETE operation.
```
felipemontoya:
```suggestion
        Note: If the role value is empty, then it must be sent as a DELETE operation.
```

eox_core/cms/views.py
```diff
@@ -129,6 +142,85 @@ def course_team(request, *args, **kwargs):
...
+
+    match_courses = []
+    course_regex = get_course_regex(input_course_regex)
+
+    courses_iter, in_process_course_actions = get_courses_accessible_to_user(request)
+    active_courses, archived_courses = get_process_courses_list(courses_iter, in_process_course_actions)
+
+    if not active_courses:
+        data = {
+            'message': 'No courses were found.',
```
felipemontoya: Do we have the piping in place to offer this as translated strings?

eox_core/cms/views.py
```diff
@@ -146,9 +238,21 @@ def get_json_from_body_request(request):
...
-    if "application/json" in request.META.get('CONTENT_TYPE', '') and request.body:
+    if 'application/json' in request.META.get('CONTENT_TYPE', '') and request.body:
         try:
             json_content = json.loads(request.body)
         except ValueError:
             return json_content
     return json_content
+
+
+def get_course_regex(input_course_regex):
```
felipemontoya: I'm not sure I understand the reason for writing this function. Is it because the original regex is not url safe and you need to do some replacements? what about doing a url_encode first then?

If the reason is different, then please clarify in the docstring.

eox_core/edxapp_wrapper/backends/courses_h_v1.py
```diff
@@ -13,3 +15,26 @@ def courses_accessible_to_user(*args, **kwargs):
...
+    """ get CourseDetails fields. """
+    course_details = CourseDetails(org=None, course_id=None, run=None)
+    all_fields = vars(course_details)
+
+    return all_fields.keys()
+
+
+def get_first_course_key():
+    """
+    Returns the course key of any course in roder to get
```
felipemontoya:
```suggestion
    Returns the course key of any course in order to get
```

webpack.config.js
```diff
@@ -1,24 +1,51 @@
 const path = require('path');
```
felipemontoya: Since we are compiling the bundle on dev for now and commiting the current bundle into the repository, we definitely need a piece on the readme about how to compile

eox_core/cms/views.py
Squirrel18: No, since we are not defining a strings catalog and translations folder. @felipemontoya

eox_core/cms/urls.py
Squirrel18: This repo has no pycodestyle enable in the CircleCI configuration file. @felipemontoya

Squirrel18:
@felipemontoya Feedback addressed. eslint is now running on the npm build, just missing to add a new make command to run those linting test, and adding to CircleCI config file.

PR #48, opened 2019-03-04
Add linting test to CircleCI.

Squirrel18:
## Description:

This PR adds linting test (pylint - eslint) to CircleCI tests.

The docker images were updated in order to use images with NodeJS and npm already installed.

## Future work:

- Check a better definition to plugin_settings function since it is defined twice, in common.py and test.py, and pylint refactor messages were disabled for now.

## Reviewers:

- [x] @diegomillan 
- [x] @felipemontoya


felipemontoya:
@Squirrel18 maybe it would be clearner to squash before merging this PR

PR #49, opened 2019-03-05
LLPA-20 - Course rerun

Squirrel18:
## Description:

This PR adds the tab called Course rerun which enables to rerun courses into a list of organizations.

## How to test:

- Install eox-core, checkout to this branch.
- Go to eox-core installed path and run: npm run build.
- Go to /admin
- Add and activate a new Waffle switch called: course_management_automation_module
- Login into Studio
- Go to eox-core/management/courses
- Select a course key of any course.
- Select the desired organizations.

## Screenshot:

image

## Reviewers:

- [x] @diegomillan 
- [ ] @felipemontoya 
- [ ] @amalbas


eox_core/cms/static/js/index.jsx
```diff
@@ -58,3 +67,9 @@ CourseManagement.propTypes = {
     request_timeout_value: PropTypes.number,
   }),
 };
+
+CourseManagement.propTypes = {
+  courseRerun: PropTypes.shape({
+    request_timeout_value: PropTypes.number,
```
diegomillan: I think we should use a standard to define attributes or variables. Whether I understand we must define a style guide for this,  maybe camelcase works better in this context. What do you think @amalbas ???

eox_core/cms/static/js/CourseRerun.jsx
```diff
@@ -0,0 +1,375 @@
...
+            isLoading: false,
+          });
+        });
+      }, requestTimeOut);
+      requestTimeOut += this.props.requestTimeOut;
+    }
+  }
+
+  handleCourseRerunResponse(response, organization) {
+    const actualSuccessTasks = this.state.completedTasks;
```
diegomillan: I think `const actualSucceededTasks` could be a better name

eox_core/cms/static/js/index.jsx
Squirrel18: Here is the only part where the props variables have underscored names because they came from of the html template and they are passed as Python context variables, could be amend to camel case names, but keeping those names like so, i think we have more separated logic between props passed from the template and child components props. @diegomillan

diegomillan: ok @Squirrel18

eox_core/cms/static/js/CourseRerun.jsx
Squirrel18: @diegomillan Addressed.

diegomillan:
Ready to merge @Squirrel18

PR #50, opened 2019-03-11
Do not raise exceptions related to empty databases

morenol:


eox_core/api/v1/permissions.py
```diff
@@ -15,12 +16,19 @@ def load_permissions():
...
-        )
+        try:
+            content_type = ContentType.objects.get_for_model(User)
+            obj, created = Permission.objects.get_or_create(  # pylint: disable=unused-variable
+                codename='can_call_eox_core',
+                name='Can access eox-core API',
+                content_type=content_type,
+            )
+        except ProgrammingError as err:
+            # Do not raise exception if it is related to database migrations
```
jfavellar90: I would add just a more detailed explanation about the issue, indicating that this code runs loading the app before running the migrations, and for that reason we're bypassing the ProgrammingError in this case, to let migrations run smoothly.

eox_core/api/v1/permissions.py
```diff
@@ -15,12 +16,19 @@ def load_permissions():
...
+            pass
+        except Exception as err: # pylint: disable=broad-except
```
felipemontoya: Do we need this, we can just not catch the other exceptions

eox_core/api/v1/permissions.py
```diff
@@ -15,12 +16,17 @@ def load_permissions():
...
-            content_type=content_type,
-        )
+        try:
+            content_type = ContentType.objects.get_for_model(User)
+            obj, created = Permission.objects.get_or_create(  # pylint: disable=unused-variable
+                codename='can_call_eox_core',
+                name='Can access eox-core API',
+                content_type=content_type,
+            )
+        except ProgrammingError as err:
```
jfavellar90: just a fix the pylint problem here (err defined but not used)

diegomillan:
Can we rebase and merge @morenol ??

morenol:
@diegomillan rebased

diegomillan:
I will merge @morenol

PR #54, opened 2019-05-16
anmrdz/support extra fields

anmrdz:
### **Description**

The purpose of this PR is to expand user creation API by having these features: 
- Enable extra_fields support in the creation, the extra fields  are  level_of_education,  gender, mailing_address, city, country, goals, and year_of_birth
- Enable extended_profile_fields, this refers to the field 'meta' in the user profile.
- Open the possibility to a refined control of the user creation flow such as :
	- Send verification email
	- Enable discussion notification emails
	- Create comment client
	- Allow third-party authentication  (social media providers)
	- Allow external authentication

The proposed Json structure would be like: 
`data = {
        'email': "address@example.org",
        'username': "Username",
        'password': "P4ssW0rd",
        'fullname': "Full Name",
        'activate': True,
        'site': request.site,
        'language_preference': 'es-419',
        'extended_profile_fields': {
            'internal_id': '12',
            ...
        },
        'year_of_birth': '1981',
        'city': 'Boston',
        'verify_account_email': True,
        'enable_comments_service': True
    }`

The way to achieve this is using the open-Edx  function [create_account_with_parms](https://github.com/eduNEXT/edunext-platform/blob/6851b666a144a9f1cd4a73ee0c51ccc3046972f8/common/djangoapps/student/views/management.py#L585)   , this allows us to leverage the maintenance of this code but will share the same issues : 

>     Issues with this code:
>     * It is non-transactional except where explicitly wrapped in atomic to
>       alleviate deadlocks and improve performance. This means failures at different places in registration
>      can leave users in inconsistent states.
>     * Third-party auth passwords are not verified. There is a comment that
>       they are unused, but it would be helpful to have a sanity check that they are sane.
>     * The user-facing text is rather unfriendly (e.g. "Username must be a
>       minimum of two characters long" rather than "Please use a username of at least two characters").
>     * Duplicate email raises a ValidationError (rather than the expected
>       AccountValidationError). Duplicate username returns an inconsistent
>       user message (i.e. "An account with the Public Username '{username}'
>       already exists." rather than "It looks like {username} belongs to an
>       existing account. Try again with a different username.") The two checks
>       occur at different places in the code; as a result, registering with
>       both a duplicate username and email raises only a ValidationError for
>       email only.

### **Work in Progress**
-  Transform scripts into unitary test inside eox-core
-  Handle format errors in expected fields
-  Display errors generated in create_account_with_parms as the request response
-  Make sure all configurations are restored after monkey patching

### **Testing**

`curl -X POST --header "Authorization: Bearer <AUTH_TOKEN> " -H "Accept: application/json" \
	http://edx.devstack.lms:18000/eox-core/api/v1/user/ --header "Content-Type: application/json" \
	--data '{"username": "eox2test", "email": "eox2test@example.com", "password": "qwerty123",
	"fullname": "JJ Smith", "gender":"f", "extended_profile_fields":{"pet":"dog", "internal_id":"12"}, "city":"Buenos Aires", "country":"AR", "level_of_education": "p"}'`

This would create a user with extra_fields and extended_profile fields. if no verify_account_email nor enable_comments_service are present these steps would be skipped. 

image

### **Reviewers**

- [ ] @felipemontoya 

- [ ] @diegomillan

### DIFF
--- a/eox_core/edxapp_wrapper/backends/users_h_v1.py
+++ b/eox_core/edxapp_wrapper/backends/users_h_v1.py
@@ -11,6 +11,7 @@ from copy import deepcopy
 from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.db import transaction
+from openedx.core.djangoapps.site_configuration import helpers as configuration_helpers
 from rest_framework.exceptions import APIException
 
 from openedx.core.djangoapps.lang_pref import (  # pylint: disable=import-error
@@ -85,16 +86,21 @@ def create_edxapp_user(*args, **kwargs):
         return None, ["Fatal: account collition with the provided: {}".format(", ".join(conflicts))]
     request = kwargs.pop("request", False)
     request.data['name'] = kwargs.pop("fullname")
+    # Get additional settings for this user
     email_validation = request.data.pop("verify_account_email", False)
     terms_of_service = request.data.pop("terms_of_service", None)
     honor_code = request.data.pop("honor_code", None)
     comments_service = request.data.pop("enable_comments_service", None)
+    extended_profile_data = request.data.pop("extended_profile_fields", {})
 
-    _original_settings_FEATURES = deepcopy(settings.FEATURES)
+    # Settings to be monkey patched to suit request
     registration_fields = getattr(settings, 'REGISTRATION_EXTRA_FIELDS', {})
+    site_configuration = configuration_helpers.get_current_site_configuration()
+    # Backup copies of original configurations
+    _original_settings_features = deepcopy(settings.FEATURES)
     _original_settings_registration_fields = deepcopy(registration_fields)
 
-    settings.FEATURES['SKIP_EMAIL_VALIDATION'] = not email_validation == True
+    settings.FEATURES['SKIP_EMAIL_VALIDATION'] = not email_validation
 
     if not comments_service:
         if 'ENABLE_DISCUSSION_SERVICE' in settings.FEATURES:
@@ -110,15 +116,20 @@ def create_edxapp_user(*args, **kwargs):
     else:
         registration_fields['honor_code'] = 'hidden'
 
+    extended_profile_fields = {
+        'extended_profile_fields': {key: '' for key in extended_profile_data}
+        }
+    site_configuration.values.update(extended_profile_fields)
+    request.data.update(extended_profile_data)
     user = create_account_with_params(request, request.data)
 
-    settings.FEATURES = _original_settings_FEATURES
+    settings.FEATURES = _original_settings_features
     registration_fields = _original_settings_registration_fields
+    #TODO restore site configuration
 
     if kwargs.pop("activate_user", False):
         user.is_active = True
         user.save()
-
     return user, errors
 
 
--- a/eox_core/edxapp_wrapper/backends/users_h_v1.py
+++ b/eox_core/edxapp_wrapper/backends/users_h_v1.py
@@ -7,6 +7,7 @@ from __future__ import absolute_import, unicode_literals
 
 import logging
 
+from copy import deepcopy
 from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.db import transaction
@@ -26,7 +27,7 @@ from student.forms import AccountCreationForm  # pylint: disable=import-error
 from student.helpers import (  # pylint: disable=import-error
     create_or_set_user_attribute_created_on_site
 )
-from student.helpers import do_create_account  # pylint: disable=import-error
+from student.views import create_account_with_params  # pylint: disable=import-error
 from student.models import CourseEnrollment  # pylint: disable=import-error
 from student.models import (UserAttribute, UserSignupSource,  # pylint: disable=import-error
                             create_comments_service_user)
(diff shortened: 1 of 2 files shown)


eox_core/edxapp_wrapper/backends/users_h_v1.py
```diff
@@ -7,9 +7,11 @@
```
felipemontoya: Are we 100% compatible with this implementation with the old one?

If we are not, then we should add a new `users_h_v2` backend and do the implementation there.
If we are, we can keep things here but we must be careful that we dont break the compatibility

eox_core/edxapp_wrapper/backends/users_h_v1.py
```diff
@@ -64,69 +66,70 @@ def create_edxapp_user(*args, **kwargs):
...
 
-    try:
-        create_comments_service_user(user)
-    except Exception:  # pylint: disable=broad-except
-        errors.append("No comments_service_user was created")
-
-    # TODO: link account with third party auth
-
-    lang_pref = kwargs.pop("language_preference", False)
-    if lang_pref:
```
felipemontoya: Can you remove a block like this? We still want to support setting the `language_preference` so unless that feature is supported on create_account_with_params then we need to keep it.

eox_core/edxapp_wrapper/backends/users_h_v1.py
```diff
@@ -64,69 +66,70 @@ def create_edxapp_user(*args, **kwargs):
...
-                user.username,
-            ))
+    extended_profile_fields = {
+        'extended_profile_fields': {key: '' for key in extended_profile_data}
+        }
+    site_configuration.values.update(extended_profile_fields)
+    request.data.update(extended_profile_data)
+    user = create_account_with_params(request, request.data)
+
+    settings.FEATURES = _original_settings_features
```
felipemontoya: the settings object is an in_memory object that outlasts the live of a particular requests and remains in the gunicorn worker for more requests.
I dont see the need to edit it, but furthermore can not allow any modifications based on a user request

eox_core/edxapp_wrapper/backends/users_h_v1.py
```diff
@@ -64,69 +66,70 @@ def create_edxapp_user(*args, **kwargs):
...
+    terms_of_service = request.data.pop("terms_of_service", None)
+    honor_code = request.data.pop("honor_code", None)
+    comments_service = request.data.pop("enable_comments_service", None)
+    extended_profile_data = request.data.pop("extended_profile_fields", {})
+
+    # Settings to be monkey patched to suit request
+    registration_fields = getattr(settings, 'REGISTRATION_EXTRA_FIELDS', {})
+    site_configuration = configuration_helpers.get_current_site_configuration()
+    # Backup copies of original configurations
+    _original_settings_features = deepcopy(settings.FEATURES)
```
felipemontoya: I see now, in my comment below I did not understand what you where doing with settings. I would much rather prefer that we use a context_manager to change the settings object instead of modifying it.

Something like what `django.test.utils.override_settings` does but only for the settings.FEATURE dict. override_settings might be too much, but still worth taking a [look](https://github.com/django/django/blob/1.11.20/django/test/utils.py#L395)

eox_core/edxapp_wrapper/backends/users_h_v1.py
```diff
@@ -64,69 +66,70 @@ def create_edxapp_user(*args, **kwargs):
...
-            preferences_api.set_user_preference(user, LANGUAGE_KEY, lang_pref)
-        except Exception:  # pylint: disable=broad-except
-            errors.append("Could not set lang preference '{} for user '{}'".format(
-                lang_pref,
-                user.username,
-            ))
+    extended_profile_fields = {
+        'extended_profile_fields': {key: '' for key in extended_profile_data}
+        }
+    site_configuration.values.update(extended_profile_fields)
```
felipemontoya: The same as above. It is not really a good idea to edit the site_configuration values in memory.
On the other hand, why should you modify `extended_profile_fields` on the site. The API should allow you to fill the values on the site, not create new on the flight

eox_core/edxapp_wrapper/backends/users_h_v1.py
```diff
@@ -64,69 +66,70 @@ def create_edxapp_user(*args, **kwargs):
...
+    request.data.update(extended_profile_data)
```
felipemontoya: If you are updating the request in place, which you should not do, then it is better to create a blank request and fill in exactly what you want it to have before sending it to the `create_account_with_params` function

eox_core/edxapp_wrapper/backends/users_h_v1.py
```diff
@@ -64,69 +66,70 @@ def create_edxapp_user(*args, **kwargs):
...
+
+    settings.FEATURES = _original_settings_features
+    registration_fields = _original_settings_registration_fields
+    #TODO restore site configuration
 
     if kwargs.pop("activate_user", False):
         user.is_active = True
         user.save()
-
-    # TODO: run conditional email sequence
```
felipemontoya: Since we have so far not implemented this, then no need to remove the #TODO mark

anmrdz:
Closing this PR because the best approach is yet not clear and it has lost priority over other features.
@felipemontoya  @diegomillan @andrey-canon

PR #62, opened 2019-06-06
More standard methods por POST & PUT

felipemontoya:
This PR refactors a bit the logic for POST and PUT to:
- make the underlying functions require a user as a parameter, removing the need for 3 different implementations of get_edxapp_user
- have a function to update/create 1 single record in the view, where the required user param is explicitly passed to the underlying function
- support the capacity to respond in bulk to the bulk responses when one or more end in errors

### DIFF
--- a/eox_core/api/v1/views.py
+++ b/eox_core/api/v1/views.py
@@ -89,6 +89,7 @@ class EdxappEnrollment(APIView):
         """
         super(EdxappEnrollment, self).__init__(*args, **kwargs)
         self.query_params = None
+        self.site = None
 
     def post(self, request, *args, **kwargs):
         """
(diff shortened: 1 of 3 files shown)


eox_core/edxapp_wrapper/backends/enrollment_h_v1.py
```diff
@@ -197,7 +220,7 @@ def enroll_on_course(course_id, *args, **kwargs):
     return enrollment, errors
 
 
-def enroll_on_program(program_uuid, *arg, **kwargs):
+def _enroll_on_program(user, program_uuid, *arg, **kwargs):
     """
     enroll user on each of the courses of a progam
```
anmrdz:
```suggestion
    enroll user on each of the courses of a program
```

eox_core/api/v1/views.py
```diff
@@ -89,25 +89,64 @@ def __init__(self, *args, **kwargs):
...
+            data = EdxappCourseEnrollmentSerializer(enrollment).data
+            if msg:
+                data["messages"] = msg
+            response_data.append(data)
+
+        return response_data
 
     def post(self, request, *args, **kwargs):
         """
         Creates the users on edxapp
```
anmrdz:
```suggestion
        Handle creation of single or bulk enrollments 
```

PR #63, opened 2019-06-07
Adding pycodestyle for code quality checks

diegomillan:
This PR adds and fixes code quality errors using pycodestyle

@felipemontoya 
@anmrdz

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -15,3 +15,8 @@ omit =
 	*/backends*
 	node_modules/*
 
+[pycodestyle]
+# error codes: https://pycodestyle.readthedocs.io/en/latest/intro.html#error-codes
+# E501: line too long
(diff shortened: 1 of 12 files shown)


setup.cfg
```diff
@@ -15,3 +15,8 @@ omit =
 	*/backends*
 	node_modules/*
 
+[pycodestyle]
+# error codes: https://pycodestyle.readthedocs.io/en/latest/intro.html#error-codes
+# E501: line too long
+ignore=E501
```
Squirrel18: Even you are ignoring line too long, I think there should be a number defined as 120. @diegomillan 
max-line-length = 120

diegomillan: But it will be ignored @Squirrel18. We should add it if we decide to remove E501

PR #66, opened 2019-06-12
Tokens should only be used in the same site as the oauth client object

felipemontoya:
With this PR the token client is compared to the host. If they don't match, then it is the same as if no authorization had happened.

This is critical for multi-tenant access.

SessionAuthorization is no longer possible unless the user is_staff = true

### DIFF
--- a/eox_core/api/v1/permissions.py
+++ b/eox_core/api/v1/permissions.py
@@ -3,7 +3,7 @@
 """
 Custom API permissions module
 """
-from rest_framework import permissions
+from rest_framework import permissions, exceptions
 from django.conf import settings
 from django.contrib.auth.models import Permission, User
 from django.contrib.contenttypes.models import ContentType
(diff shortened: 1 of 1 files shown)


eox_core/api/v1/permissions.py
```diff
@@ -32,10 +32,22 @@ def load_permissions():
...
+        try:
+            allowed_for_site = request.user.is_staff or request.get_host() in request.auth.client.url
+        except Exception:  # pylint: disable=broad-except
+            allowed_for_site = False
+
+        if not allowed_for_site:
+            # If we get here either someone is using a token created on one site in a different site
+            # Or there was a missconfiguration of the oauth client.
+            # To prevent leaking important information we return the most basic message
+            raise exceptions.NotAuthenticated(detail="Invalid token")
```
Squirrel18: Maybe, you can raise this exception on line 45, since 'allowed_for_site' is only used to raise this exception. @felipemontoya

eox_core/api/v1/permissions.py
```diff
@@ -32,10 +32,22 @@ def load_permissions():
 class EoxCoreAPIPermission(permissions.BasePermission):
     """
     Defines a custom permissions to access eox-core API
+    This permissions makes sure that a token is created with the client credentias of the same site is being used on
```
Squirrel18: Please, end the line comments with a period. @felipemontoya

eox_core/api/v1/permissions.py
felipemontoya: I had that first, but when I added the user.is_staff check the line's logic became hard to follow

eox_core/api/v1/permissions.py
```diff
@@ -32,10 +32,22 @@ def load_permissions():
 class EoxCoreAPIPermission(permissions.BasePermission):
     """
     Defines a custom permissions to access eox-core API
+    This permissions makes sure that a token is created with the client credentias of the same site is being used on.
```
anmrdz:
```suggestion
    These permissions make sure that a token is created with the client credentials of the same site is being used on.
```

eox_core/api/v1/permissions.py
```diff
@@ -32,10 +32,22 @@ def load_permissions():
...
     """
 
     def has_permission(self, request, view):
         """
         To grant accees, checks if the request user either can call eox-core API or if it's an admin user.
```
anmrdz:
```suggestion
        To grant access, checks if the requesting user either can call eox-core API or if it's an admin user.
```

PR #67, opened 2019-06-14
Feedback from 1.0.0

felipemontoya:
This PR solves the errors we found on V1.0.0 once it was up in production.

- Site is available for POST in enrollments
- No missing migrations
- Manifest with includes so no -e is needed in installation

### DIFF
--- a/eox_core/api/v1/views.py
+++ b/eox_core/api/v1/views.py
@@ -54,6 +54,15 @@ class UserQueryMixin(object):
         self.query_params = None
         self.site = None
 
+    def initial(self, request, *args, **kwargs):
+        """
+        Loads the site into the object for every kind of request
+        """
+        super(UserQueryMixin, self).initial(request, *args, **kwargs)
+
+        if hasattr(request, 'site'):
+            self.site = request.site
+
     def get_query_params(self, request):
         """
         Utility to read the query params in a forgiving way
@@ -64,9 +73,6 @@ class UserQueryMixin(object):
             query_params = request.data
 
         self.query_params = query_params
-
-        if hasattr(request, 'site'):
-            self.site = request.site
 
         return query_params
 
@@ -191,9 +197,6 @@ class EdxappEnrollment(UserQueryMixin, APIView):
         """
         Update enrollments on edxapp
         """
-        if hasattr(request, 'site'):
-            self.site = request.site
-
         data = request.data
         return EdxappEnrollment.prepare_multiresponse(data, self.single_enrollment_update)
 
--- a/eox_core/api/v1/views.py
+++ b/eox_core/api/v1/views.py
@@ -56,7 +56,7 @@ class UserQueryMixin(object):
 
     def initial(self, request, *args, **kwargs):
         """
-        Loads the site into the object for every kind of request
+        Loads the site into the object for every kind of request.
         """
         super(UserQueryMixin, self).initial(request, *args, **kwargs)
 
--- a/eox_core/edxapp_wrapper/backends/certificates_h_v1_test.py
+++ b/eox_core/edxapp_wrapper/backends/certificates_h_v1_test.py
@@ -2,14 +2,14 @@
 Test backend to get GenerateCertificates Model.
 """
 
-from django.db import models
+from django.contrib.auth.models import Permission
 
 
 def get_generated_certificate():
-    """ get test GeneratedCertificate model. """
+    """
+    get test GeneratedCertificate model.
 
-    class CertificatesDummyModel(models.Model):
-        """ Dummy model for testing. """
-        pass
-
-    return CertificatesDummyModel
+    We return any django model that already exists so that
+    django-filters is happy and no migrations are created.
+    """
+    return Permission
--- a/eox_core/edxapp_wrapper/backends/certificates_h_v1_test.py
+++ b/eox_core/edxapp_wrapper/backends/certificates_h_v1_test.py
@@ -7,7 +7,7 @@ from django.contrib.auth.models import Permission
 
 def get_generated_certificate():
     """
-    get test GeneratedCertificate model.
+    Get test GeneratedCertificate model.
 
     We return any django model that already exists so that
     django-filters is happy and no migrations are created.
--- a/eox_core/edxapp_wrapper/backends/users_h_v1.py
+++ b/eox_core/edxapp_wrapper/backends/users_h_v1.py
@@ -190,25 +190,25 @@ class FetchUserSiteSources(object):
 
     @classmethod
     def get_enabled_source_methods(cls):
-        """ brings the array of methods to check if an user belongs to a site """
+        """ Brings the array of methods to check if an user belongs to a site. """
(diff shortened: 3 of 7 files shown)


eox_core/edxapp_wrapper/backends/certificates_h_v1_test.py
```diff
@@ -2,14 +2,14 @@
...
 
-    class CertificatesDummyModel(models.Model):
-        """ Dummy model for testing. """
-        pass
-
-    return CertificatesDummyModel
+    We return any django model that already exists so that
+    django-filters is happy and no migrations are created.
+    """
+    return Permission
```
jfavellar90: I did not understand why you can pass any model here. I suppose you chose a django built-in one by convenience

eox_core/edxapp_wrapper/backends/users_h_v1_test.py
```diff
@@ -2,14 +2,14 @@
...
 
-    class CourseEnrollmentDummyModel(models.Model):
-        """ Dummy model for testing. """
-        pass
-
-    return CourseEnrollmentDummyModel
+    We return any django model that already exists so that
+    django-filters is happy and no migrations are created.
+    """
+    return Permission
```
jfavellar90: Same question above

eox_core/api/v1/views.py
```diff
@@ -54,6 +54,15 @@ def __init__(self, *args, **kwargs):
         self.query_params = None
         self.site = None
 
+    def initial(self, request, *args, **kwargs):
```
jfavellar90: who calls this method?

eox_core/api/v1/views.py
```diff
@@ -54,6 +54,15 @@ def __init__(self, *args, **kwargs):
...
+        """
+        Loads the site into the object for every kind of request
```
Squirrel18: Please, add a period at the end of the line. @felipemontoya

eox_core/api/v1/views.py
```diff
@@ -54,6 +54,15 @@ def __init__(self, *args, **kwargs):
...
+        """
+        super(UserQueryMixin, self).initial(request, *args, **kwargs)
+
+        if hasattr(request, 'site'):
+            self.site = request.site
```
Squirrel18: If request doesn't have site, what happens next? @felipemontoya

eox_core/edxapp_wrapper/backends/users_h_v1_test.py
```diff
@@ -2,14 +2,14 @@
...
 """
 
-from django.db import models
+from django.contrib.auth.models import Permission
 
 
 def get_course_enrollment():
-    """ get Test CourseEnrollment model """
+    """
+    get Test CourseEnrollment model
```
Squirrel18: Capitalize Get, please @felipemontoya

eox_core/edxapp_wrapper/backends/users_h_v1.py
```diff
@@ -197,7 +197,9 @@ def get_enabled_source_methods(cls):
     @staticmethod
     def fetch_from_created_on_site_prop(user, domain):
         """ fetch option """
```
Squirrel18: Capitalize fetch, please @felipemontoya and add a period at the end of the line.

eox_core/edxapp_wrapper/backends/certificates_h_v1_test.py
```diff
@@ -2,14 +2,14 @@
...
 """
 
-from django.db import models
+from django.contrib.auth.models import Permission
 
 
 def get_generated_certificate():
-    """ get test GeneratedCertificate model. """
+    """
+    get test GeneratedCertificate model.
```
Squirrel18: Capitalize Get, please @felipemontoya

eox_core/edxapp_wrapper/backends/users_h_v1_test.py
Squirrel18: And add a period at the end of the line.

eox_core/edxapp_wrapper/backends/certificates_h_v1_test.py
felipemontoya: Exactly. One that already exists. I can pass a new model, but that creates a migration and then tests of the lms fail because `makemigrations` now creates output which the test dislike

eox_core/api/v1/views.py
felipemontoya: the dispatch method in DRF

eox_core/edxapp_wrapper/backends/users_h_v1.py
```diff
@@ -197,7 +197,9 @@ def get_enabled_source_methods(cls):
...
-        return UserAttribute.get_user_attribute(user, 'created_on_site') == domain
+        if domain:
```
Squirrel18: This could be if not domain: return False, and leave the UserAttribute out of the if, perhaps in the future should do something else. @felipemontoya

eox_core/api/v1/views.py
felipemontoya: If the request does not have a site, then None is the default value. It happens in many tests

eox_core/edxapp_wrapper/backends/users_h_v1.py
felipemontoya: The same thing as I am doing here. Is that some prefered pythonic pattern or something?

Squirrel18: It's not a PEP8 standard, but it will help when the function grows and will keep the code more readable, since, you are catching the incorrect doamin value fisrt of all, and then, you can put the rest of code outside of the if. https://docs.python-guide.org/writing/style/#returning-values

eox_core/api/v1/views.py
diegomillan: Just a minor comment:

I did not understand how `super()` was working in this context. Here two references to check:
https://rhettinger.wordpress.com/2011/05/26/super-considered-super/
https://stackoverflow.com/questions/27954695/what-is-a-sibling-class-in-python

`super(UserQueryMixin, self).initial(request, *args, **kwargs)` is a call for [DRF initial method defined in views ](https://github.com/encode/django-rest-framework/blob/afb678433b7dc068213e6e8a359ad1f6fff05b0d/rest_framework/views.py#L387) since it's the sibling class.

cc: @felipemontoya @jfavellar90 @Squirrel18

eox_core/edxapp_wrapper/backends/users_h_v1.py
felipemontoya: Done

eox_core/api/v1/views.py
felipemontoya: @diegomillan just checked using ipdb and the DRF method is getting called. So it is working as it should. I'm closing this then

diegomillan: Yes, it is working @felipemontoya. I just wanted to explain the reason why,  if anyone wants to check. 
LGTM this is approved.

PR #70, opened 2019-07-09
Add two openedx models to the Django admin

morenol:


eox_core/admin.py
```diff
@@ -2,4 +2,34 @@
 """ Admin.py """
 from __future__ import unicode_literals
 
-# Register your models here.
+from django.contrib import admin
+from django.contrib.admin.sites import AlreadyRegistered
+
+
+from student.models import UserSignupSource, LoginFailures  # pylint: disable=import-error
```
jfavellar90: I'm wondering if this import import should be wrapped on the edxapp_wrapper folder, as this is our communication bridge with the platform.

morenol: Done

PR #84, opened 2020-03-20
Adding sentry integration

jfavellar90:
This PR adds Sentry integration support through eox-core installation with extras_require option.

The plugin installation must be as fllows:
`pip install git+https://github.com/eduNEXT/eox-core.git@v2.7.0#egg=eox-core==2.7.0[sentry]`

This will install the sentry SDK to enable the configuration. The readme file was updated to include details on how to enable sentry integration.

### DIFF
--- a/eox_core/settings/common.py
+++ b/eox_core/settings/common.py
@@ -46,3 +46,6 @@ def plugin_settings(settings):
             'fetch_from_created_on_site_prop',
             'fetch_from_user_signup_source',
         ]
+
+    # Sentry Integration
+    settings.EOX_CORE_SENTRY_INTEGRATION_DSN = None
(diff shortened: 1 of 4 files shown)


eox_core/settings/common.py
```diff
@@ -46,3 +46,7 @@ def plugin_settings(settings):
             'fetch_from_created_on_site_prop',
             'fetch_from_user_signup_source',
         ]
+
+    # Sentry Integration
+    settings.EOX_CORE_SENTRY_INTEGRATION_ENABLED = False
+    settings.EOX_CORE_SENTRY_INTEGRATION_DSN = "fake_dsn"
```
morenol: I think that instead of two separate configurations we can have one and use it at this way:

None if we want to disable that or "string" if we want to enable that.

jfavellar90: Agree, the change was applied.

PR #86, opened 2020-04-06
Do not send NotFoundError to sentry

morenol:


eox_core/settings/aws.py
```diff
@@ -120,7 +120,24 @@ def plugin_settings(settings):  # pylint: disable=function-redefined
         settings.EOX_CORE_SENTRY_INTEGRATION_DSN
     )
     if sentry_sdk is not None and sentry_integration_dsn is not None:
+        def before_send(event, hint):
```
felipemontoya, line 123: For me it makes sense to define this function in a different file.
And also to make the list of ignored errors a configurable value

morenol:
@felipemontoya can you take a look?

eox_core/settings/common.py
```diff
@@ -49,3 +49,7 @@ def plugin_settings(settings):
 
     # Sentry Integration
     settings.EOX_CORE_SENTRY_INTEGRATION_DSN = None
+    settings.EOX_CORE_SENTRY_IGNORED_ERRORS = [
+        'xmodule.exceptions.NotFoundError',
```
felipemontoya, line 53: Can we put this as an example on the sentry.py module docstring and remove it from here.

Then we can add it only for the SAAS on EOX_CORE_SENTRY_IGNORED_ERRORS

morenol: I added that to the README

PR #88, opened 2020-04-07
Enable celery integration for sentry

morenol:
Upgrade [sentry-sdk](https://github.com/getsentry/sentry-python/blob/master/CHANGES.md) version and enable [celery integration](https://docs.sentry.io/platforms/python/celery/)

https://stackoverflow.com/a/60947153

### DIFF
--- a/eox_core/settings/aws.py
+++ b/eox_core/settings/aws.py
@@ -5,7 +5,7 @@ from .common import *  # pylint: disable=wildcard-import, unused-wildcard-import
 
 try:
     import sentry_sdk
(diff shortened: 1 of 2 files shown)


morenol:
Right now, we dont have any problems if we set that as sentry-sdk, but if a new version of sentry-sdk has breaking changes this will be broken.

I think that in the thing that we need to do is to make this repo use o[ep-18 practices](https://open-edx-proposals.readthedocs.io/en/latest/oep-0018-bp-python-dependencies.html#id9)

PR #89, opened 2020-04-09
Use 150 as username max length in Ironwood

morenol:
In ironwood the max length allowed for a username is 150. 

I thought that we could make use of EOX_CORE_USERS_BACKEND instead of add a new setting for this.

I changed the default of settings.EOX_CORE_USERS_BACKEND to the ironwood version


felipemontoya:
Using the users backend is a good idea. If more cases like this come up, we could create a constant in the backend instead of a function, but we would need to change the interface definitions. For now I think this is fine.

PR #90, opened 2020-04-15
Use username max length defined in edx-platform

morenol:


felipemontoya:
With this in mind we would also not need the users_i backend, right?

morenol:
> With this in mind we would also not need the users_i backend, right?

No, let me know if do you think that it would be better to remove that.

felipemontoya:
I think so, yeah. Why leave code creating the I backend when it only
imports the H backend? Let's remove it for clarity

felipemontoya:
:+1:

PR #91, opened 2020-04-16
Adding the configurable backend

felipemontoya:
This PR adds a completely configurable OpenId Connect based backend for third party auth


eox_core/social_tpa_backends.py
```diff
@@ -0,0 +1,35 @@
+"""
+Extensions to the regular defined third party auth backends
+"""
+import logging
+
+from django.conf import settings
+from social_core.backends.open_id_connect import OpenIdConnectAuth
```
morenol, line 7: This is a new dependency to eox-core, maybe it is a good idea to add that to the requirements

felipemontoya: Im thinking better to import it conditionally

morenol:
LGTM, we only need to fix the quality errors

PR #92, opened 2020-04-17
Add social_core[openidconnect] as extra requirement

morenol:
Add extra dependency to social_core[openidconnect] in order to install python-jose if needed with 

eox-core[tpa]


felipemontoya:
No need to put a version on the line?

morenol:
> No need to put a version on the line?

I think that no, at this way we will use the edx-platform version.

PR #94, opened 2020-05-01
Filtering sentry exceptions based on traceback text regex matches

jfavellar90:
This PR implements a callable to ignore specific exceptions to be sent to sentry.io service. The motivation behind this PR is the fact that Sentry dashboard gets a bit noisy with lots of exception events that are not relevant since they're not truly errors.

The exceptions can be ignored based on exception class match or last traceback text regex match

@diegomillan 
@morenol 
@felipemontoya

### DIFF
--- a/eox_core/integrations/sentry.py
+++ b/eox_core/integrations/sentry.py
@@ -3,7 +3,9 @@ This file implements utils used for sentry integration.
 
 See: https://github.com/eduNEXT/eox-core#integrations-with-third-party-services
 """
+import re
 import importlib
+import six
 
 from django.conf import settings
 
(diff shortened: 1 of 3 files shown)


jfavellar90:
@felipemontoya I measured the time it takes to run with just one defined rule and it was 0,8 milliseconds approx, so I don't think it's problematic unless we have a bunch of rules

felipemontoya:
I'm sure we will have a lot of rules, but I was asking more about when in the request life cycle does this happen. Have we sent the response already or are we waiting for this sentry part to complete before we answer the http?

jfavellar90:
@felipemontoya we have to wait for this to get run before sending a response. I don't know if we can find a way to run this snippet in async mode since this is nor relevant for the response

jfavellar90:
Sentry protects by itself the calls to before_send method. If an exception is raised in the method, the event is dropped. Anyways I added a try-except wrapping the evaluation of every ignoring rule to get further information about a possible error.

PR #96, opened 2020-06-11
Test using the python version used in juniper. Use circleci matrix

morenol:
Use circleci matrix and test using python version of Juniper.

I am using python3.5.8 and python2.7.16 because the docker images python2.7-node and python3.5-node have node12.0.0 and it is currently failing (I think that we have to solve that for juniper since in juniper the node version is 12.0.0 and in ironwood it is V8.9.3). 

These images are using node version: 10.16.3 (just like the image that we were using).

I am also using the stretch debian base image since it is the image compatible with ubuntu16.04.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,26 +3,35 @@
 # Check https://circleci.com/docs/2.0/language-python/ for more details
 #
 
-version: 2
+version: 2.1
 workflows:
-  version: 2
+  version: 2.1
   test:
     jobs:
-      - test-3.6
-      - test-2.7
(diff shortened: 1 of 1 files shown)


felipemontoya:
If we are already doing this, why not include python 3.8 as it will be in koa?

morenol:
> If we are already doing this, why not include python 3.8 as it will be in koa?

The current problem with that is that the circleci docker images with python3.8 have node12.0 and that version is currently failing in eox-core(but that is something that we need to fix for Juniper)

PR #105, opened 2020-08-17
Added patch method to edit edxapp users

mariajgrimaldi:
This PR adds patch method to the new view EdxappUserUpdater used to update users fields

This is a response to the necessity of updating the password and the field is_active using eox-core.

### DIFF
--- a/eox_core/api/v1/serializers.py
+++ b/eox_core/api/v1/serializers.py
@@ -4,6 +4,7 @@ API v1 serializers.
 # pylint: disable=abstract-method
 from __future__ import absolute_import, unicode_literals
 
+from django.conf import settings
 from rest_framework import serializers
 
 from eox_core.edxapp_wrapper.coursekey import get_valid_course_key, validate_org
@@ -11,8 +12,13 @@ from eox_core.edxapp_wrapper.enrollments import check_edxapp_enrollment_is_valid
 from eox_core.edxapp_wrapper.users import (
     check_edxapp_account_conflicts,
     get_user_read_only_serializer,
+    get_user_signup_source,
     get_username_max_length,
 )
+
+UserSignupSource = get_user_signup_source()  # pylint: disable=invalid-name
+
+MAX_SIGNUP_SOURCES_ALLOWED = 1
 
 USERNAME_MAX_LENGTH = get_username_max_length()
 
@@ -69,6 +75,54 @@ class EdxappUserSerializer(serializers.Serializer):
         if conflicts:
             raise serializers.ValidationError("Account already exists with the provided: {}".format(", ".join(conflicts)))
         return attrs
+
+
+class WrittableEdxappUserSerializer(EdxappUserSerializer):
+    """
+    Handles the serialization of the user data required to update an edxapp user.
+    """
+    password = serializers.CharField(
+        style={'input_type': 'password'},
+    )
+    is_active = serializers.BooleanField()
+
+    def validate(self, attrs):
+        """
+        When an update is being made, then it checks that the user:
+            - Is not staff or superuser.
+            - Has just one signup source.
+            - The field being updated is a safe field.
+        """
+        # If at least one of these conditions is true, then the user can't be updated.
+        safe_fields = getattr(settings, "EOX_CORE_USER_UPDATE_SAFE_FIELDS", [])
+
+        for attr in attrs:
+            if attr not in safe_fields:
+                raise serializers.ValidationError({"detail": "You are not allowed to update {}.".format(attr)})
+
+        if self.instance.is_staff or self.instance.is_superuser:
+            raise serializers.ValidationError({"detail": "You can't update users with roles like staff or superuser."})
+
+        if UserSignupSource.objects.filter(user__email=self.instance.email).count() > MAX_SIGNUP_SOURCES_ALLOWED:
+            raise serializers.ValidationError({"detail": "You can't update users with more than one sign up source."})
+
+        return attrs
+
+    def update(self, instance, validated_data):
+        """
+        Update method for safe fields.
+        """
+        for key, value in validated_data.items():
+
+            if key == "password":
+                instance.set_password(value)
+            else:
+                setattr(instance, key, value)
+
+        if validated_data:
+            instance.save()
+
+        return instance
 
 
 class EdxappUserQuerySerializer(EdxappUserSerializer):
--- a/eox_core/api/v1/tests/test_users.py
+++ b/eox_core/api/v1/tests/test_users.py
@@ -1,0 +1,77 @@
+"""
+Test module for users viewset.
+"""
+from django.contrib.auth.models import User
+from django.test import TestCase
+from django.urls import reverse
+from mock import patch
+from rest_framework import status
+from rest_framework.test import APIClient
+
+
+class UsersUpdaterAPITest(TestCase):
+    """Test class for users viewset."""
+
+    patch_permissions = patch('eox_core.api.v1.permissions.EoxCoreAPIPermission.has_permission', return_value=True)
+
+    def setUp(self):
+        """Setup method for test class."""
+        self.user = User(username="test", email="test@example.com", password="testtest")
+        self.client = APIClient()
+        self.url = reverse("eox-api:eox-api:edxapp-user-updater")
+        self.client.force_authenticate(user=self.user)
+
+    @patch_permissions
+    @patch('eox_core.api.v1.serializers.UserSignupSource')
+    @patch('eox_core.api.v1.views.get_edxapp_user')
+    @patch('eox_core.api.v1.views.EdxappUserReadOnlySerializer')
+    def test_update_success(self, user_serializer, get_edxapp_user, signup_source, _):
+        """Used to test updating safe fields of an edxapp user."""
+        update_data = {
+            "email": "test",
+            "fullname": "test",
+        }
+        user_serializer.return_value.data = {}
+        get_edxapp_user.return_value = self.user
+        signup_source.objects.filter.return_value.count.return_value = 1
+
+        response = self.client.patch(self.url, data=update_data, format="json")
+
+        self.assertEqual(status.HTTP_200_OK, response.status_code)
+
+    @patch_permissions
+    @patch('eox_core.api.v1.serializers.UserSignupSource')
+    @patch('eox_core.api.v1.views.get_edxapp_user')
+    def test_update_bad_sign_up_source(self, get_edxapp_user, signup_source, _):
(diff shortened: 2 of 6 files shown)


eox_core/api/v1/serializers.py
```diff
@@ -4,6 +4,7 @@
 # pylint: disable=abstract-method
 from __future__ import absolute_import, unicode_literals
 
+from django.utils.translation import ugettext as _
```
andrey-canon, line 7: Does this plugin have translations ?

eox_core/api/v1/serializers.py
```diff
@@ -16,6 +17,8 @@
 
 USERNAME_MAX_LENGTH = get_username_max_length()
 
+USER_UPDATE_SAFE_FIELDS = ("fullname", "password", "is_active",)
```
andrey-canon, line 20:
```suggestion
USER_UPDATE_SAFE_FIELDS = ("fullname", "password", "is_active")
```

eox_core/api/v1/serializers.py
```diff
@@ -68,8 +71,27 @@ def validate(self, attrs):
...
+                if key not in USER_UPDATE_SAFE_FIELDS:
+                    raise serializers.ValidationError({key: _("Field not allowed to update.")})
+
         return attrs
 
+    def update(self, instance, validated_data):
+        """
+        Update method for safe fields.
+        """
+        for key in validated_data:
```
andrey-canon, line 87: it's better items() ?

eox_core/api/v1/views.py
```diff
@@ -137,6 +137,21 @@ def get(self, request, *args, **kwargs):
 
         return Response(response_data)
 
+    def patch(self, request, *args, **kwargs):
+        """
```
andrey-canon, line 141: could you add an example with the url type of request and data which are required

eox_core/api/v1/views.py
```diff
@@ -137,6 +137,21 @@ def get(self, request, *args, **kwargs):
...
+        Partially update an user from edxapp. Not all the fields can be updated, just the ones thought as `safe`.
+        """
+        query = self.get_user_query(request)
```
andrey-canon, line 144: is this getting the information from the query params  ?

mariajgrimaldi: Yes. This way we can use the method without adding new URLs and we can use the username or email through the query params. Do you think that is OK or is it better do define new paths like user/<username> or user/<email>?

eox_core/api/v1/views.py
mariajgrimaldi: Yes, I'll do it.

eox_core/api/v1/serializers.py
mariajgrimaldi: you're right

eox_core/api/v1/views.py
andrey-canon: this is very odd, if you don't want to make new urls  you can get that information from the body

mariajgrimaldi: I decided to add a new URL just for /user/<EMAIL>/. What do you think?

eox_core/api/v1/tests/test_users.py
```diff
@@ -0,0 +1,97 @@
...
+class UsersAPITest(TestCase):
+    """Test class for users viewset."""
+
+    patch_permissions = patch('eox_core.api.v1.permissions.EoxCoreAPIPermission.has_permission', return_value=True)
+
+    def setUp(self):
+        """Setup method for test class."""
+        self.user = User(username="test", email="test@example.com", password="testtest")
+        self.client = APIClient()
+        self.url = "/api/v1/user/"
```
andrey-canon, line 19: reverse

eox_core/api/v1/tests/test_users.py
```diff
@@ -0,0 +1,97 @@
...
+        user_serializer.return_value.data = {}
+        get_edxapp_user.return_value = self.user
+        signup_source.objects.filter.return_value.count.return_value = 1
+
+        response = self.client.patch(self.url, data=update_data, format="json")
+
+        self.assertEqual(response.status_code, 200)
+
+    @patch_permissions
+    @patch('eox_core.api.v1.serializers.USER_UPDATE_SAFE_FIELDS')
```
andrey-canon, line 41: you don't need to patch constants

eox_core/api/v1/tests/test_users.py
```diff
@@ -0,0 +1,97 @@
...
+            "username": "test",
+            "password": "test",
+        }
+        user_serializer.return_value.data = {}
+        get_edxapp_user.return_value = self.user
+        signup_source.objects.filter.return_value.count.return_value = 1
+
+        response = self.client.patch(self.url, data=update_data, format="json")
+
+        self.assertEqual(response.status_code, 200)
```
andrey-canon, line 38: the convention says that first the expected value and second the result, use status from drf instead of 200

eox_core/api/v1/tests/test_users.py
```diff
@@ -0,0 +1,97 @@
...
+            "username": "test",
+            "password": "test",
+        }
+        get_edxapp_user.return_value = self.user
+        signup_source.objects.filter.return_value.count.return_value = 1
+        safe_fields.return_value = ("fullname")
+
+        response = self.client.patch(self.url, data=update_data, format="json")
+
+        self.assertEqual(response.status_code, 400)
```
andrey-canon, line 59: same comments

eox_core/api/v1/tests/test_users.py
```diff
@@ -0,0 +1,97 @@
...
+        update_data = {
+            "username": "test",
+            "password": "test",
+        }
+        get_edxapp_user.return_value = user
+        signup_source.objects.filter.return_value.count.return_value = 1
+
+        response = self.client.patch(self.url, data=update_data, format="json")
+
+        self.assertEqual(response.status_code, 400)
```
andrey-canon, line 95: same comments

andrey-canon (approved):

eox_core/api/v1/serializers.py
```diff
@@ -11,9 +11,14 @@
 from eox_core.edxapp_wrapper.users import (
     check_edxapp_account_conflicts,
     get_user_read_only_serializer,
+    get_user_signup_source,
     get_username_max_length,
 )
 
+UserSignupSource = get_user_signup_source()  # pylint: disable=invalid-name
+
+MAX_SIGNUP_SOURCES_TO_UPDATE = 1
```
felipemontoya, line 20: This constant is a bit confusing. I thought that you where going to update the signups sources.
What about calling it: MAX_SIGNUP_SOURCES_ALLOWED?

eox_core/api/v1/urls.py
```diff
@@ -10,3 +11,8 @@
     url(r'^pre-enrollment/$', views.EdxappPreEnrollment.as_view(), name='edxapp-pre-enrollment'),
     url(r'^userinfo/$', views.UserInfo.as_view(), name='edxapp-userinfo'),
 ]
+
+if getattr(settings, "EOX_CORE_UPDATE_USERS_ENABLE", None):
```
felipemontoya, line 15: Would you consider changing the name to EOX_CORE_ENABLE_UPDATE_USERS?

eox_core/api/v1/serializers.py
```diff
@@ -71,6 +76,48 @@ def validate(self, attrs):
...
+        style={'input_type': 'password'},
+    )
+    is_active = serializers.BooleanField()
+
+    def validate(self, attrs):
+        """
+        When an update is being made, then it checks that the user:
+            - Is not staff or superuser.
+            - Has just one signup source.
+            - The field being updated is a safe field.
```
felipemontoya, line 93: I dont see this validation that the field is safe. Is it just that the key exists in this serializer?
Could I by mistake update the username using this?

eox_core/api/v1/views.py
```diff
@@ -138,6 +139,51 @@ def get(self, request, *args, **kwargs):
...
+            "email": data.pop("email", None),
+        }
+        query = self.get_user_query(request, query_params=query_params)
+        user = get_edxapp_user(**query)
+
+        serializer = WrittableEdxappUserSerializer(user, data=data, partial=True)
+        serializer.is_valid(raise_exception=True)
+        serializer.save()
+
+        admin_fields = getattr(settings, 'ACCOUNT_VISIBILITY_CONFIGURATION', {}).get('admin_fields', {})
```
felipemontoya, line 182: I am not familiar with this setting. Is it from this plugin? From django, from the edx-platform core?
What are you expecting this admin_fields to contain?

mariajgrimaldi: I followed how get users returned the data:
https://github.com/eduNEXT/eox-core/blob/36823821634e49f34abf799e4bbb180b75f8f84d/eox_core/api/v1/views.py#L130

should I changed it?

eox_core/api/v1/serializers.py
mariajgrimaldi: agree

eox_core/api/v1/urls.py
mariajgrimaldi: alright, that's better

eox_core/api/v1/serializers.py
mariajgrimaldi: That was what I had before. I got confused when we talked about adding the new serializer. 
I'm going to add the following:
SAFE_FIELDS_TO_UPDATE = ["is_active", "password", "fullname"]
should I add another field?

mariajgrimaldi:
I want to highlight something about the update process: 
To be able to update a user, we must pass the user email in the request body.

Why the email and not the username?
Well, 

1. This feature is going to be used by an application that provides only the user's email.
2. When updating, is easier to let the user know what happened with their account using their email.

eox_core/api/v1/views.py
felipemontoya: I don't know. Perhaps it makes sense to refactor that into a method that we understand better, but I think we have dragged this long enough.

Here is the commit that added this line. Just for context as it does not add much info:
https://github.com/eduNEXT/eox-core/commit/e9b6dd225d1b99fdf58ca61de8954366c75e597c

felipemontoya (approved):
I'll leave my approve, but I do think we should make the fields a setting

eox_core/api/v1/serializers.py
mariajgrimaldi: ok nice, I'll add the setting and put those fields as defaults

PR #108, opened 2020-09-17
Add django 2.2 test

andrey-canon:
## Description 
The main object is to test in django 2.2, the support for django 2.2 will be include in other pr


morenol (approved):
LGTM, please squash.

I think that with also need to check if we need to gmfix the issues with node12

PR #112, opened 2020-10-06
Update install and dependencies update instructions.

MoisesGSalas:
There are a few instructions that need updating.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -36,7 +36,7 @@ Usage
 
 Installation on Open edX Devstack
 =================================
-- Install the Ironwood version of the `Open edX devstack`_
+- Install either the Ironwood or Juniper version of the `Open edX devstack`_
 
 - Clone the git repo:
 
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -53,21 +53,14 @@ Installation on Open edX Devstack
 
 	cd ~/Documents/eoxstack/devstack  # Change for your devstack path (if you are using devstack)
 	make lms-shell  # Enter the devstack machine (or server where lms process lives)
-	sudo su edxapp -s /bin/bash  # if you are using devstack you need to use edxapp user
-	source /edx/app/edxapp/venvs/edxapp/bin/activate
 	cd /edx/src/edxapp/eox-core
 	pip install -e .
 
 Development
```
morenol, line 59: I think that we can change this title with something more related to the information that is below.

MoisesGSalas: Something like ''Remove unnecessary installation steps and update dependency management instructions." would do?

morenol (approved):

PR #116, opened 2020-10-19
Updating permissions.py to accept dot and dop tokens

vad33r:


eox_core/api/v1/permissions.py
```diff
@@ -40,7 +40,11 @@ def has_permission(self, request, view):
         To grant access, checks if the requesting user either can call eox-core API or if it's an admin user.
         """
         try:
-            allowed_for_site = request.user.is_staff or request.get_host() in request.auth.client.url
+            if request.user.is_staff:
+                allowed_for_site = request.user.is_staff
+            else:
+                allowed_for_site = request.auth.application.redirect_uri_allowed(request.build_absolute_uri('/')) or \
```
andrey-canon, line 46: This always raises an exception, doesn't it ? when the request is with dop

eox_core/api/v1/permissions.py
```diff
@@ -40,7 +40,11 @@ def has_permission(self, request, view):
         To grant access, checks if the requesting user either can call eox-core API or if it's an admin user.
         """
         try:
-            allowed_for_site = request.user.is_staff or request.get_host() in request.auth.client.url
+            if request.user.is_staff:
+                allowed_for_site = request.user.is_staff
```
felipemontoya, line 44: This first part of the if is just confusing. If the is in fact staff we are making the whole function take a different path. My suggestion would be to completely remove from the logic by looking that up first and returning.
If the user is not staff, then you can do the auth.application or auth.client checks each in a try..catch block

felipemontoya (changes requested):
Please remember to add the unit tests for this changeset

eox_core/api/v1/permissions.py
vad33r: No, when the request is with dop tries with:
`request.get_host() in request.auth.client.url`

mariajgrimaldi:
I know the title of the PR is self-explanatory but can you add a description? That way if someone returns to this PR they can have more context.

eox_core/api/v1/permissions.py
```diff
@@ -39,8 +39,14 @@ def has_permission(self, request, view):
         """
         To grant access, checks if the requesting user either can call eox-core API or if it's an admin user.
         """
+        if request.user.is_staff:
+            return request.user.is_staff
         try:
-            allowed_for_site = request.user.is_staff or request.get_host() in request.auth.client.url
+            allowed_for_site = request.get_host() in request.auth.client.url
```
andrey-canon, line 45: if this is True, the line 49 will raise the exception and do this False ...

eox_core/api/v1/permissions.py
```diff
@@ -39,8 +39,17 @@ def has_permission(self, request, view):
...
         try:
-            allowed_for_site = request.user.is_staff or request.get_host() in request.auth.client.url
+            allowed_for_site = request.auth.application.redirect_uri_allowed(request.build_absolute_uri('/'))
+            return request.user.has_perm('auth.can_call_eox_core')
+        except Exception:  # pylint: disable=broad-except
+            allowed_for_site = False
+
+        try:
+            allowed_for_site = request.get_host() in request.auth.client.url
+            return request.user.has_perm('auth.can_call_eox_core')
```
andrey-canon, line 52:
```suggestion
            if hasattr(request.auth, 'application'):
                allowed_for_site = request.auth.application.redirect_uri_allowed(request.build_absolute_uri('/'))
            else:
                allowed_for_site = request.get_host() in request.auth.client.url
```

vad33r: Ok, I didn't know what I could do `if hasattr(request.auth, 'application'):` ty

andrey-canon (commented):
@vad33r  did you test these last changes ??

vad33r:
Yes, i tested them

andrey-canon (approved):

PR #118, opened 2020-11-03
Fix enrollment import error in Juniper

mariajgrimaldi:
This PR fixes the following error in Juniper:

Screenshot from 2020-11-03 16-46-38

This error was raised because in juniper enrollments API is located in /openedx/core/djangoapps/enrollments  and not in /common/djangoapps/enrollment. This was done in edx/edx-platform//pull/20702

- [x] Tested in Juniper
- [x] Tested in Ironwood

### DIFF
--- a/eox_core/edxapp_wrapper/backends/enrollment_h_v1.py
+++ b/eox_core/edxapp_wrapper/backends/enrollment_h_v1.py
@@ -11,8 +11,6 @@ import logging
 
 from course_modes.models import CourseMode
 from django.contrib.auth.models import User
-from enrollment import api
-from enrollment.errors import CourseEnrollmentExistsError, CourseModeNotFoundError
 from opaque_keys import InvalidKeyError
 from opaque_keys.edx.keys import CourseKey
 from openedx.core.djangoapps.content.course_overviews.models import CourseOverview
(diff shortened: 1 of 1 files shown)


felipemontoya:
@MoisesGSalas fyi

MoisesGSalas:
This seems to have fixed the issue I've been having. Thank you very much, @mariajgrimaldi.

eox_core/edxapp_wrapper/backends/enrollment_h_v1.py
```diff
@@ -25,6 +23,19 @@
 from eox_core.edxapp_wrapper.coursekey import get_valid_course_key, validate_org
 from eox_core.edxapp_wrapper.users import check_edxapp_account_conflicts
 
+try:
```
andrey-canon, line 26: Imo we should do this as a new backend `eox_core/edxapp_wrapper/backends/enrollment_j_v1.py`

mariajgrimaldi: even considering that we are only fixing the import?

felipemontoya: Since the change is backwards compatible with hawhtorn I would say that changing the _h_v1 backend is good enough

felipemontoya (approved):

PR #119, opened 2020-11-04
Fix django-filters in Juniper

mariajgrimaldi:
**Description and justification**

This PR fixes users'/course enrollment/ generated certificates/proctored exams filter error, caused by django-filter forward compatibility issue. Also, makes it backward compatible.

The main issue that we have here is that in django-filter v1.0.4 -version used in Ironwood- there's a keyword `name` that was deprecated after v2.0.0 -Juniper uses v2.2.0-, then or we use the deprecated keyword `name` or we use the replacement `field_name`. 

Solution: 

Change name to field_name and change Django-filter to a version != than 1.0.4 when using python 2.7. Version 1.1.0 supports name and field_name keyword: https://github.com/carltongibson/django-filter/blob/1.1.0/django_filters/filters.py#L92

### DIFF
--- a/eox_core/api/data/v1/filters.py
+++ b/eox_core/api/data/v1/filters.py
@@ -30,16 +30,16 @@ class UserFilter(BaseDataApiFilter):
     date_joined = django_filters.DateTimeFromToRangeFilter()
 
     # Filtering by user profile fields
-    name = django_filters.CharFilter(name="profile__name", lookup_expr="icontains")
-    language = django_filters.CharFilter(name="profile__language", lookup_expr="iexact")
-    year_of_birth = django_filters.RangeFilter(name="profile__year_of_birth")
-    gender = django_filters.CharFilter(name="profile__gender", lookup_expr="iexact")
-    mailing_address = django_filters.CharFilter(name="profile__mailing_address", lookup_expr="iexact")
-    city = django_filters.CharFilter(name="profile__city", lookup_expr="icontains")
-    country = django_filters.CharFilter(name="profile__country", lookup_expr="icontains")
+    name = django_filters.CharFilter(field_name="profile__name", lookup_expr="icontains")
+    language = django_filters.CharFilter(field_name="profile__language", lookup_expr="iexact")
+    year_of_birth = django_filters.RangeFilter(field_name="profile__year_of_birth")
+    gender = django_filters.CharFilter(field_name="profile__gender", lookup_expr="iexact")
+    mailing_address = django_filters.CharFilter(field_name="profile__mailing_address", lookup_expr="iexact")
+    city = django_filters.CharFilter(field_name="profile__city", lookup_expr="icontains")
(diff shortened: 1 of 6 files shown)


eox_core/api/data/v1/filters.py
```diff
@@ -10,6 +12,43 @@
 from eox_core.edxapp_wrapper.users import get_course_enrollment
 
 
+def check_python_version():
```
morenol, line 15: I think that we should use six for this.

See: https://six.readthedocs.io/#six.PY2

eox_core/settings/common.py
```diff
@@ -19,7 +19,7 @@ def plugin_settings(settings):
     Defines eox-core settings when app is used as a plugin to edx-platform.
     See: https://github.com/edx/edx-platform/blob/master/openedx/core/djangoapps/plugins/README.rst
     """
-    settings.EOX_CORE_USERS_BACKEND = "eox_core.edxapp_wrapper.backends.users_j_v1"
+    settings.EOX_CORE_USERS_BACKEND = "eox_core.edxapp_wrapper.backends.users_h_v1"
```
andrey-canon, line 22: revert this please

eox_core/api/data/v1/filters.py
```diff
@@ -199,4 +199,4 @@ class Meta(object):
             'course_id',
             'exam_name'
 
-        ]
+        ]
```
andrey-canon, line 202: a line is missing

andrey-canon (commented):
You have to force the django-filter update, maybe writing django-filter>1.0.4 in the base.in is enough

eox_core/api/data/v1/filters.py
mariajgrimaldi: thanks but I changed the solution!

Following what @andrey-canon suggested:

Given the problem I described, we could update django-filters in Ironwood from 1.0.4 to 1.1.0, this particular version supports name (raises just a deprecation warning) and field_name [reference](https://github.com/carltongibson/django-filter/blob/1.1.0/django_filters/filters.py#L92) What do you think?

requirements/base.txt
```diff
@@ -11,7 +11,7 @@ celery==3.1.25            # via -c requirements/constraints.txt, -r requirements
 certifi==2020.6.20        # via requests
 chardet==3.0.4            # via requests
 django-crum==0.7.7        # via edx-proctoring
-django-filter==1.0.4      # via -c requirements/constraints.txt, -r requirements/base.in
+django-filter==1.1.0      # via -c requirements/constraints.txt, -r requirements/base.in
```
andrey-canon, line 14: please use the command `make upgrade`  to update this file

mariajgrimaldi: with python 3.5 or 2.7?

andrey-canon: 2.7 maybe it's necessary the constraint to 1.1.0

andrey-canon (approved):
LGTM

PR #121, opened 2020-12-16
Remove api user info permission.

andrey-canon:
## Description
That permission doesn't make a lot of sense since this view returns the current user information and doesn't make anything else, and due to request.user is set by BearerAuthentication or SessionAuthentication the information returned always will belong to the current user.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,6 +12,10 @@ Change Log
 Unreleased
 ----------
 
+Removed
+~~~~~~~
+* EoxCoreAPIPermission from UserInfo APIView
+
 [3.2.0] - 2020-11-18
 --------------------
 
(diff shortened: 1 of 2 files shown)


felipemontoya:
I don't necessarily agree that having this permission is not relevant, but for a better understanding, Id'l like to ask. What do we gain by removing this permission?
Where has this permission been an issue?

andrey-canon:
> I don't necessarily agree that having this permission is not relevant, but for a better understanding, Id'l like to ask. What do we gain by removing this permission?
> Where has this permission been an issue?

@felipemontoya 
Right now the permission is not an issue, this view https://github.com/edx-unsupported/edx-oauth2-provider/blob/master/edx_oauth2_provider/views.py#L217 was deprecated in juniper, so another view is required to return the user info hence if we remove the permission this view could fit, if there is another option let me know

mariajgrimaldi (approved):

felipemontoya:
Do you mean removed or only deprecated in juniper?

andrey-canon:
remove

felipemontoya:
I came from basecamp from reading the context and for the moment I think its ok if use this view.

felipemontoya (approved):

PR #122, opened 2020-12-17
Allow inactive users in UserInfo api view

andrey-canon:
## Description
The current version only allows active users, so this fails if the user is new and doesn't make the validation, so this change adds a new backend for inactive users and edit the authentication classes for the UserInfo api view.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,6 +12,11 @@ Change Log
 Unreleased
 ----------
 
+Added
+~~~~~
+* BearerAuthenticationAllowInactiveUser backend.
+* IsAuthenticated permission to UserInfo api view.
+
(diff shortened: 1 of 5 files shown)


mariajgrimaldi (approved):

felipemontoya:
I am not completely sold on this approach, but at this time we don't have anything better to offer.

In the long run, I believe we would be better suited by a real OIDC provider in a separate plugin.

felipemontoya (approved):
Lets go with this until we can figure out the ID provider

andrey-canon:
After another review I found a view which already returns the user info without special permission so this pr and the change applied in https://github.com/eduNEXT/eox-core/pull/121 are not necessary.

PR #123, opened 2020-12-17
Add EoxCoreAPIPermission to UserInfo APIView

andrey-canon:
## Description
revert changes of https://github.com/eduNEXT/eox-core/pull/121 since the view my_user_info in https://github.com/edx/edx-platform/blob/master/lms/djangoapps/mobile_api/users/views.py  already returns the necessary data without special permissions.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,6 +12,10 @@ Change Log
 Unreleased
 ----------
 
+Added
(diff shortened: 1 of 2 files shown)


mariajgrimaldi (approved):
-.-

felipemontoya (approved):

PR #125, opened 2021-01-07
Add swagger support for documentation.

MoisesGSalas:
The current documentation is no easily discoverable, with this PR with can make our documentation public and of easy access from any site that uses eox-core. The main benefits are:

1. Live documentation, can perform interactive requests.
2. Swagger it's [already](https://github.com/edx/edx-platform/pull/21207) on the platform.
3. Can be used to produce static documentation.

(2021/01/11 edit): Some test requirements were not correctly upgraded on #126, so I also added them.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,6 +12,9 @@ Change Log
 Unreleased
 ----------
 
+* Add swagger support
+* Improve internal documentation
+
 [3.4.0] - 2020-12-16
 --------------------
 
--- a/eox_core/api/v1/serializers.py
+++ b/eox_core/api/v1/serializers.py
@@ -4,6 +4,8 @@ API v1 serializers.
 # pylint: disable=abstract-method
 from __future__ import absolute_import, unicode_literals
 
+from collections import OrderedDict
+
 from django.conf import settings
 from rest_framework import serializers
 
@@ -171,8 +173,24 @@ class EdxappCourseEnrollmentSerializer(serializers.Serializer):
     username = serializers.CharField(max_length=USERNAME_MAX_LENGTH, default=None, source='user')
     is_active = serializers.BooleanField(default=True)
     mode = serializers.CharField(max_length=100)
-    enrollment_attributes = EdxappEnrollmentAttributeSerializer(many=True, default=[])
+    enrollment_attributes = EdxappEnrollmentAttributeSerializer(many=True, required=False)
     course_id = EdxappValidatedCourseIDField()
+
+    class Meta:
+        """
+        Add extra details for swagger
+        """
+        swagger_schema_fields = {
+            "example": OrderedDict(
+                [
+                    ("username", "johndoe"),
+                    ("is_active", True),
+                    ("mode", "audit"),
+                    ("enrollment_attributes", []),
+                    ("course_id", "course-v1:edX+DemoX+Demo_Course")
+                ]
+            ),
+        }
 
     def validate(self, attrs):
         """
--- a/eox_core/api/v1/views.py
+++ b/eox_core/api/v1/views.py
@@ -6,6 +6,7 @@ from __future__ import absolute_import, unicode_literals
 
 import logging
 
+import edx_api_doc_tools as apidocs
 from django.conf import settings
 from django.contrib.sites.shortcuts import get_current_site
 from django.utils import six
@@ -43,6 +44,7 @@ class UserQueryMixin:
     """
     Provides tools to create user queries
     """
+
     def __init__(self, *args, **kwargs):
         """
         Defines instance attributes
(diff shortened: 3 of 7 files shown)


eox_core/api/v1/serializers.py
```diff
@@ -171,9 +172,25 @@ class EdxappCourseEnrollmentSerializer(serializers.Serializer):
     username = serializers.CharField(max_length=USERNAME_MAX_LENGTH, default=None, source='user')
     is_active = serializers.BooleanField(default=True)
     mode = serializers.CharField(max_length=100)
-    enrollment_attributes = EdxappEnrollmentAttributeSerializer(many=True, default=[])
+    enrollment_attributes = EdxappEnrollmentAttributeSerializer(many=True, required=False)
```
magajh, line 175: why was the default deleted here?

eox_core/api/v1/urls.py
```diff
@@ -18,3 +20,11 @@
     urlpatterns += [
         url(r'^update-user/$', views.EdxappUserUpdater.as_view(), name='edxapp-user-updater'),
     ]
+
+urlpatterns += [
+    url(
```
magajh, line 25: Is there a specific reason why this URL should be appended to urlpatterns here? If not I think it could be added where urlpatterns is defined, maybe leaving a blank line between this api-docs URL and the rest.

eox_core/api/v1/api_schema.py
```diff
@@ -0,0 +1,59 @@
...
+    Schema generator for eox-core.
+
+    Define an exclusive base path to perform requests for each endpoint and a
+    specific security definition using oauth without overwritting project wide
+    settings.
+    """
+
+    def get_schema(self, request=None, public=False):
+        schema = super().get_schema(request, public)
+        schema.base_path = '/eox-core/api/v1/'
```
magajh, line 26: Since '/eox-core/api/v1/' is being used more than once maybe you should consider declaring a constant at the beginning of the file.

eox_core/api/v1/api_schema.py
```diff
@@ -0,0 +1,59 @@
...
+                'flow': 'application',
+                'tokenUrl': '{}/oauth2/access_token/'.format(settings.LMS_ROOT_URL),
+                'type': 'oauth2',
+            },
+        }
+        security_definitions = SwaggerDict.as_odict(security_definitions)
+        return security_definitions
+
+
+api_urls = [  # pylint: disable=invalid-name
```
magajh, line 41: More URLs could be added here later, I think we should move this to another file
What do you think?

eox_core/api/v1/serializers.py
MoisesGSalas: drf-yasg creates an openapi object based on the type of fields of the serializer, because `enrollment_attributes` is a nested serializer and we are using a default it tries to explore the `EdxappEnrollmentAttributeSerializer` fields against an empty list resulting on a traceback error.

eox_core/api/v1/api_schema.py
MoisesGSalas: This one is actually an oversight, the second time I use that path it's not doing anything so I'm going to delete it.

eox_core/api/v1/urls.py
MoisesGSalas: I was thinking of having them in a separate set in case we want to disable the view with a setting or something like that.

eox_core/api/v1/api_schema.py
magajh: ok

eox_core/api/v1/api_schema.py
MoisesGSalas: I'm no quite sure how to handle this one. For now the idea was to add paths as the documentation for each resource is completed. Eventually we'll end up having a subset of `urls.py`, I'm not too much of a fan having those duplicates.

eox_core/api/v1/serializers.py
magajh: Ok. If you tested this and it does not affect the expected behavior of the request then lgtm

eox_core/api/v1/urls.py
magajh: In that case, we could do something similar to what was done with "EOX_CORE_ENABLE_UPDATE_USERS", in the meantime, in my opinion, we should add it where urlpatterns is defined

eox_core/api/v1/views.py
```diff
@@ -191,59 +195,202 @@ class EdxappEnrollment(UserQueryMixin, APIView):
...
         """
-        Get enrollments on edxapp
+        Retrieves enrollment information given a user and a course_id
+
+        **Example Requests**
+
+            GET /eox-core/api/v1/enrollment/?username=johndoe&
+            course_id=course-v1:edX+DemoX+Demo_Course
+
+            Request data: {
```
magajh, line 384: in this request data, according to the documentation, you could pass either the email or username. Maybe that could be specified here somehow?

MoisesGSalas: Where would that be? That request data is more like a visual aid because reading the whole string can be a little bit difficult. In that example I'm performing a request with `username`, I didn't want to add more examples because it gets a bit too long and I added a small explanation on the description of the parameters.

magajh: Yes, you are right. No need to add it here since this is just an example

eox_core/api/v1/urls.py
MoisesGSalas: Done.

morenol:
It seems that this generates and endpoint in /eox-core/api/v1/api-docs/

What do you think, if we instead use:
 
/eox-core/api-docs/

magajh (approved):

magajh (approved):

MoisesGSalas:
@morenol, that change seems fine to me, the only consideration would be if there was ever a v2 of the api, maybe we would like to have docs for each version?

morenol:
>  @morenol, that change seems fine to me, the only consideration would be if there was ever a v2 of the api, maybe we would like to have docs for each version? 

I think that all docs for both versions would be in /eox-core/api-docs/

MoisesGSalas:
@morenol, that change seems fine to me, the only consideration would be if there was ever a v2 of the api, maybe we would like to have docs for each version? 

> > @morenol, that change seems fine to me, the only consideration would be if there was ever a v2 of the api, maybe we would like to have docs for each version?
> 
> I think that all docs for both versions would be in /eox-core/api-docs/

In that case I would go and make that change.

eox_core/api/v1/urls.py
```diff
@@ -1,5 +1,6 @@
 """ urls.py """
 
+
```
magajh, line 3: Delete this please

morenol:
@MoisesGSalas could you add an entry in the Changelog? It can be in the unreleased section

morenol:
@MoisesGSalas  I tried to install this in a clean Juniper Devstack and it is failing due to the default value of api-doc-tools installed there. Could you add to requirements/base.in the constraint (edx-api-doc-tools>=1.4.0)

eox_core/api_schema.py
```diff
@@ -0,0 +1,58 @@
...
+    Schema generator for eox-core.
+
+    Define an exclusive base path to perform requests for each endpoint and a
+    specific security definition using oauth without overwritting project wide
+    settings.
+    """
+
+    def get_schema(self, request=None, public=False):
+        schema = super().get_schema(request, public)
+        schema.base_path = '/eox-core/api/v1/'
```
morenol, line 26: Can you change this with

`eox-core/api/?

so it can include future versions?

MoisesGSalas: If I do that, the requests are made against `/eox-core/api/<endpoint>/` instead of `/eox-core/api/v1/<endpoint>/`. I didn't find a way to fix that.

MoisesGSalas: Even if I do that it drops the prefix.

morenol: Ok, then I think that we can keep it at this way for now

MoisesGSalas:
@morenol:
> @MoisesGSalas I tried to install this in a clean Juniper Devstack and it is failing due to the default value of api-doc-tools installed there. Could you add to requirements/base.in the constraint (edx-api-doc-tools>=1.4.0)

> @MoisesGSalas could you add an entry in the Changelog? It can be in the unreleased section

Done.

magajh (approved):

CHANGELOG.rst
```diff
@@ -12,6 +12,9 @@ Change Log
 Unreleased
 ----------
 
+* Add swagger support
```
morenol, line 15: This is in somehow a breaking change, maybe this should be specified here or once we will create the next feature

MoisesGSalas: I thought the actual breaking change was made in #126.

morenol: Yes, but AFAIK in that PR there are not changes that are incompatible with the previous version of api-doc-tools

morenol (approved):

PR #127, opened 2021-01-20
Added pipeline step to avoid disconnection from TPA provider

mariajgrimaldi:
This PR adds a pipeline function that can avoid disconnection from a TPA provider through TPA Configuration settings.

    For example:
        To avoid disconnection from SAML, add the following to `Other config str` in your SAMLConfiguration:

        "BACKEND_OPTIONS": { "disableDisconnectPipeline":true },

        Now, to avoid disconnection from an Oauth2Provider, add the same setting to `Other settings` in your
        Oauth2Provider.

It's recommended to place this function at the beginning of the pipeline.

### DIFF
--- a/eox_core/pipeline.py
+++ b/eox_core/pipeline.py
@@ -7,6 +7,11 @@ from django.db.models.signals import post_save
 
 from eox_core.edxapp_wrapper.users import get_user_profile
 
+try:
+    from social_core.exceptions import NotAllowedToDisconnect
+except ImportError:
+    NotAllowedToDisconnect = object
+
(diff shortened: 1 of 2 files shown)


morenol (commented):
@mariajgrimaldi  Looks great! 

Could you add a test for the case in which  disableDisconnectPipeline is false or undefined?

mariajgrimaldi:
> @mariajgrimaldi Looks great!
> 
> Could you add a test for the case in which disableDisconnectPipeline is false or undefined?

Great idea! 

Done :)

morenol (approved):
nice!

PR #128, opened 2021-01-21
Add Grades Endpoint

MoisesGSalas:
This PR introduces a new endpoint `eox-core/api/v1/grade` to retrieve Grade data of a user from a given course.

### DIFF
--- a/eox_core/api/v1/tests/test_grades.py
+++ b/eox_core/api/v1/tests/test_grades.py
@@ -1,0 +1,184 @@
+"""
+Test module for the Grades API
+"""
+from django.contrib.auth.models import User
+from django.test import TestCase
+from django.urls import reverse
+from mock import MagicMock, patch
+from rest_framework import status
+from rest_framework.test import APIClient
+
+from eox_core.api.v1.views import EdxappGrade
+
+
+class TestGradesAPI(TestCase):
+    """ Tests for the grades endpoints """
+
+    patch_permissions = patch(
+        "eox_core.api.v1.permissions.EoxCoreAPIPermission.has_permission",
+        return_value=True,
+    )
+
+    def setUp(self):
+        """ setup """
+        self.api_user = User(1, "test@example.com", "test")
+        self.client = APIClient()
+        self.client.force_authenticate(user=self.api_user)
+        self.url = reverse("eox-api:eox-api:edxapp-grade")
+
+    @patch("eox_core.api.v1.views.get_courseware_courses")
+    @patch("eox_core.api.v1.views.get_valid_course_key")
+    @patch("eox_core.api.v1.views.get_course_grade_factory")
+    @patch("eox_core.api.v1.views.get_enrollment")
+    @patch("eox_core.api.v1.views.get_edxapp_user")
+    @patch_permissions
+    def test_get_grade_no_detail_no_policy(  # pylint: disable=too-many-arguments
+            self,
+            _,
+            get_edxapp_user,
+            get_enrollment,
+            grade_factory,
+            get_valid_course_key,
+            get_courseware_courses,
+    ):
+        """Test that the GET method works with default parameters"""
+
+        get_edxapp_user.return_value.username = "test"
+        get_enrollment.return_value = None, None
+        grade_factory.return_value.return_value.read.return_value.percent = 0.5
+        grade_factory.return_value.return_value.read.return_value.subsection_grades = []
+        get_courseware_courses.return_value.get_course_by_id.return_value.grading_policy = (
+            {}
+        )
+        params = {
+            "course_id": "course-v1:org+course+run",
+            "username": "test",
+        }
+        expected_response = {"earned_grade": 0.5}
+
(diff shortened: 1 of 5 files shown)


morenol:
@MoisesGSalas looks good! Great job! Looking forward for the tests and the api-docs integration!

eox_core/api/v1/views.py
```diff
@@ -665,6 +669,95 @@ def log(self, desc, data, exception=None):
...
+
+        grade_factory = get_course_grade_factory()
+        course_key = get_valid_course_key(course_id)
+        course = get_courseware_courses().get_course_by_id(course_key)
+        course_grade = grade_factory().read(user, course)
+        response = {"earned_grade": course_grade.percent}
+
+        if detailed:
+            breakdown = self._section_breakdown(course_grade.subsection_grades)
+            response['section_brekdown'] = breakdown
```
mariajgrimaldi, line 723: can you use just double quotes or single, not both

eox_core/api_schema.py
```diff
@@ -40,6 +40,7 @@ def get_security_definitions(self):
 
 api_urls = [  # pylint: disable=invalid-name
     url(r'^enrollment/$', views.EdxappEnrollment.as_view(), name='edxapp-enrollment'),
+    url(r'^grade/$', views.EdxappGrade.as_view(), name='edxapp-grade'),
```
mariajgrimaldi, line 43: is this for doc?

mariajgrimaldi (commented):
Looks great

eox_core/api/v1/views.py
morenol: I suggest you to run [black](https://github.com/psf/black) on this file

eox_core/api_schema.py
MoisesGSalas: Yes.

eox_core/api/v1/views.py
MoisesGSalas: Done. But I let the imports as they were before to avoid isort warnings. If we end up adopting black in the future maybe we should look into our style configs?

eox_core/api/v1/tests/test_grades.py
```diff
@@ -0,0 +1,150 @@
...
+            self,
+            _,
+            get_edxapp_user,
+            get_enrollment,
+            grade_factory,
+            get_valid_course_key,
+            get_courseware_courses,
+    ):
+        """Test that the GET method works with section details and grading policy"""
+
```
mariajgrimaldi, line 111: you can remove this line (and the others)

eox_core/api/v1/tests/test_grades.py
```diff
@@ -0,0 +1,150 @@
...
+        response = self.client.get(self.url, params)
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(response.data, expected_response)
+
+    @patch("eox_core.api.v1.views.get_enrollment")
+    @patch("eox_core.api.v1.views.get_edxapp_user")
+    @patch_permissions
+    def test_get_grade_input_validation(self, _, get_edxapp_user, get_enrollment):
+        """ Test that the GET method requires some parameters """
```
mariajgrimaldi, line 140: can you be a bit more specific with the description?

eox_core/api/v1/tests/test_grades.py
```diff
@@ -0,0 +1,150 @@
...
+            grade_factory,
+            get_valid_course_key,
+            get_courseware_courses,
+    ):
+        """Test that the GET method works with section details and grading policy"""
+
+        get_edxapp_user.return_value.username = "test"
+        get_enrollment.return_value = None, None
+        grade_factory.return_value.return_value.read.return_value.percent = 0.5
+        grade_factory.return_value.return_value.read.return_value.subsection_grades = {}
```
mariajgrimaldi, line 115: why don't you add a real subsection_grades instead of {}? It could be useful as an example

MoisesGSalas: Mostly because then I would have to mock a subsection grade object and the test becomes a little too big. One thing I could do is make a test for the helper function `_section_breakdown` and check if it does return what is supposed to. This test already checks that `_section_breakdown` returns an empty list when passed an empty dict and the view adds it to the response when using `detailed=true`.

eox_core/api/v1/tests/test_grades.py
MoisesGSalas: Sure thing.

eox_core/api/v1/tests/test_grades.py
MoisesGSalas: Are you sure? those are the docstrings, if I do that I would get linting warnings unless I disable them.

eox_core/api/v1/tests/test_grades.py
mariajgrimaldi: You're right. Adding a test to `_section_breakdown` would be nice!

morenol (approved):

PR #130, opened 2021-01-26
Added pipeline step to check if user's information matches

mariajgrimaldi:
This PR checks when linking to a TPA provider that the user's information from the LMS (ie. email, etc) matches with the information returned by the TPA provider (ie. email, etc). If the information does not match, then an Exception will be raised.

This step will avoid linking/authenticating accounts with inconsistencies with their user information.

### DIFF
--- a/eox_core/pipeline.py
+++ b/eox_core/pipeline.py
@@ -8,8 +8,9 @@ from django.db.models.signals import post_save
 from eox_core.edxapp_wrapper.users import get_user_profile
 
 try:
-    from social_core.exceptions import NotAllowedToDisconnect
+    from social_core.exceptions import AuthFailed, NotAllowedToDisconnect
 except ImportError:
+    AuthFailed = object
     NotAllowedToDisconnect = object
 
 LOG = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


eox_core/pipeline.py
```diff
@@ -87,4 +88,31 @@ def check_disconnect_pipeline_enabled(backend, *args, **kwargs):
...
+
+        "BACKEND_OPTIONS": { "matchUserFields":["email",...] }
+
+    It's recommended to place this step after the user is available (ie. after ensure_user_information step) and
+    before user association (ie. before user_association step)
+    """
+    if not user:
+        return
+
+    defined_fields = backend.setting("BACKEND_OPTIONS", {}).get("matchUserFields", []) if backend else []
```
MoisesGSalas, line 112: Why not something like:
```suggestion
    if backend:
        defined_fields = backend.setting("BACKEND_OPTIONS", {}).get("matchUserFields", [])
    else:
        return
```
?

mariajgrimaldi: What about:

```
if not (user and backend):
    return
```

Thank you for the suggestion!

MoisesGSalas: Looks good.

eox_core/pipeline.py
```diff
@@ -87,4 +88,31 @@ def check_disconnect_pipeline_enabled(backend, *args, **kwargs):
...
+    """
+    if not (user and backend):
+        return
+
+    defined_fields = backend.setting("BACKEND_OPTIONS", {}).get("assertUserInformationMatchFields", [])
+
+    for field in defined_fields:
+
+        if str(details.get(field, "")) != str(getattr(user, field)):
+            LOG.exception("Credentials not allowed: user (%s) information does not match.", user.username)
```
morenol, line 117: I think that it is worth to mention in the log which was the field that does not match

mariajgrimaldi: Yes! I'll change the log.

eox_core/pipeline.py
```diff
@@ -87,4 +88,35 @@ def check_disconnect_pipeline_enabled(backend, *args, **kwargs):
...
+    before user association (ie. before user_association step)
+    """
+    if not (user and backend):
+        return
+
+    defined_fields = backend.setting("BACKEND_OPTIONS", {}).get("assertUserInformationMatchFields", [])
+
+    for field in defined_fields:
+
+        if str(details.get(field, "")) != str(getattr(user, field)):
```
morenol, line 116: Since getattr doesn't have the same default that .get, I think that this will fail if in both the user and the provider the field is undefined, is this the expected behavior?

mariajgrimaldi: I missed that. Thanks!

morenol (approved):

mariajgrimaldi:
FYI @cocococosti @DonatoB-eduNEXT 

The documentation can be found [here](https://docs.edunext.co/en/latest/internal/technical-procedures/escalation/TPA_pipeline_steps.html)

PR #131, opened 2021-01-26
Add documentation for the Grades API.

MoisesGSalas:
This PR includes the swagger documentation for Grades API added on #128.

### DIFF
--- a/eox_core/api/v1/serializers.py
+++ b/eox_core/api/v1/serializers.py
@@ -257,7 +257,7 @@ class EdxappGradingPolicySerializer(serializers.Serializer):
     """
     Serializes the course grading policy
     """
-    grader = EdxappGradingRawPolicySerializer(many=True, required=False, source="GRADER")
+    grader = EdxappGradingRawPolicySerializer(many=True, source="GRADER")
     grade_cutoffs = serializers.DictField(child=serializers.FloatField(), source="GRADE_CUTOFFS")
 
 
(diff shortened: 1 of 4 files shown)


eox_core/api/v1/views.py
```diff
@@ -695,20 +695,74 @@ class EdxappGrade(UserQueryMixin, APIView):
...
 
             Request data: {
               "username": "johndoe",
               "course_id": "course-v1:edX+DemoX+Demo_Course",
             }
 
+        **Response details**
+
+        - `earned_grade`: Final user score for the course.
+        - `section_breakdown`: Details of each grade subsection.
```
morenol, line 748: Can we put in this place that this is optional?  would that break the format expected by drf-yasg?

MoisesGSalas: This is part is just markdown, so adding a little text saying is optional would be no problem.

eox_core/api/v1/serializers.py
```diff
@@ -257,7 +257,7 @@ class EdxappGradingPolicySerializer(serializers.Serializer):
     """
     Serializes the course grading policy
     """
-    grader = EdxappGradingRawPolicySerializer(many=True, required=False, source="GRADER")
+    grader = EdxappGradingRawPolicySerializer(many=True, source="GRADER")
```
mariajgrimaldi, line 260: why is it required now?

MoisesGSalas: Because I realized that when you are retrieving a grading policy it always includes the `grader` and `grade_cutoffs` keys. The whole object `grading_policy` is what is optional in the response.

mariajgrimaldi (approved):

morenol (approved):

PR #132, opened 2021-01-28
Add integration tests for the Grades API

MoisesGSalas:
This PR adds integration tests for the Grades API introduced on #128

### DIFF
--- a/eox_core/api/v1/views.py
+++ b/eox_core/api/v1/views.py
@@ -790,10 +790,10 @@ class EdxappGrade(UserQueryMixin, APIView):
         course_grade = grade_factory().read(user, course)
         response = {"earned_grade": course_grade.percent}
 
-        if detailed:
+        if detailed in ("True", "true", "on", "1"):
             breakdown = self._section_breakdown(course_grade.subsection_grades)
             response["section_breakdown"] = breakdown
-        if grading_policy:
+        if grading_policy in ("True", "true", "on", "1"):
             response["grading_policy"] = course.grading_policy
 
         return Response(EdxappGradeSerializer(response).data)
--- a/eox_core/tests/integration/test_grade_integration.py
+++ b/eox_core/tests/integration/test_grade_integration.py
@@ -1,0 +1,179 @@
+"""
+Integration test suite.
+
+This suite performs multiple http requests to guarantee
+that the Grade API is behaving as expected on a live server.
+"""
+import json
+from os import environ
+
+import pytest
+import requests
+from django.test import TestCase
+from rest_framework import status
+
+
+@pytest.mark.skipif(not environ.get("TEST_INTEGRATION"), reason="Run only explicitly")
+class TesteGradeIntegration(TestCase):  # pragma: no cover
+    """Test suite"""
+
+    data = {}
+
+    @classmethod
+    def setUpClass(cls):
+        with open("eox_core/tests/integration/test_data") as file_obj:
+            cls.data = json.load(file_obj)
+        cls.data["endpoint"] = "eox-core/api/v1/grade/"
+        site1_data = {
+            "client_id": cls.data["site1_data"]["client_id"],
+            "client_secret": cls.data["site1_data"]["client_secret"],
+            "grant_type": "client_credentials",
+        }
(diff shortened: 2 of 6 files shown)


eox_core/tests/integration/test_grade_integration.py
```diff
@@ -0,0 +1,175 @@
...
+            "Host": site1_data["host"],
+        }
+        request_url = "{}/{}".format(site1_data["base_url"], self.data["endpoint"])
+
+        response = requests.get(request_url, data=data, headers=headers)
+        response_content = response.json()
+
+        self.assertEqual(response.status_code, 200)
+        self.assertIn("earned_grade", response_content)
+        self.assertIn("section_breakdown", response_content)
```
morenol, line 102: Maybe we should assert that the response does not contains the grade_policy string. The same applies for the other tests

MoisesGSalas: You're right, I'll change that.

eox_core/tests/integration/test_grade_integration.py
```diff
@@ -0,0 +1,175 @@
...
+        headers = {
+            "Authorization": "Bearer {}".format(site1_data["token"]),
+            "Host": site1_data["host"],
+        }
+        request_url = "{}/{}".format(site1_data["base_url"], self.data["endpoint"])
+
+        response = requests.get(request_url, data=data, headers=headers)
+        response_content = response.json()
+
+        self.assertEqual(response.status_code, 200)
```
mariajgrimaldi, line 76: instead of "200" try to use status.HTTP_200_OK

eox_core/tests/integration/test_grade_integration.py
```diff
@@ -0,0 +1,175 @@
...
+        Get grades info from a user enrolled on a course. Include detailed info
+        for each graded subsection.
+        """
+        site1_data = self.data["site1_data"]
+        data = {
+            "email": site1_data["user_email"],
+            "course_id": site1_data["course"]["id"],
+            "detailed": True,
+        }
+        headers = {
```
mariajgrimaldi, line 91: if you are using the same headers I think you could move them to setUp

eox_core/tests/integration/test_grade_integration.py
```diff
@@ -0,0 +1,175 @@
...
+            "grading_policy": True,
+        }
+        headers = {
+            "Authorization": "Bearer {}".format(site2_data["token"]),
+            "Host": site2_data["host"],
+        }
+        request_url = "{}/{}".format(site2_data["base_url"], self.data["endpoint"])
+
+        response = requests.get(request_url, data=data, headers=headers)
+        self.assertEqual(response.status_code, 404)
```
mariajgrimaldi, line 153: Same here, use status not found

eox_core/tests/integration/test_grade_integration.py
MoisesGSalas: Not always, some tests run headers for site 1 and others for site 2.

eox_core/tests/integration/test_grade_integration.py
MoisesGSalas: On it.

eox_core/api/v1/views.py
```diff
@@ -790,10 +790,10 @@ def get(self, request):
         course_grade = grade_factory().read(user, course)
         response = {"earned_grade": course_grade.percent}
 
-        if detailed:
+        if detailed in ("True", "true", "on", "1"):
```
morenol, line 793: Is this something that we support in the other endpoints?  Does the api docs  says that we can enabled them with "on"?

MoisesGSalas: The enrollment endpoint uses a `force` flag and uses a serializer to validate the input. There seems to be [several valid inputs](https://github.com/encode/django-rest-framework/blob/19655edbf782aa1fbdd7f8cd56ff9e0b7786ad3c/rest_framework/fields.py#L705). It also checks for valid false values and this line does not, so any other value will be taken as false. What would be your suggestion?

morenol (approved):

PR #133, opened 2021-02-03
feat(tpa): create pipeline step to ensure creation of users with usable password

morenol:
At the creation of new users through some TPA providers, some of them are created with an unusable password, a user with an unusable password cannot login if the common.djangoapps.third_party.pipeline.set_logged_in_cookies step is enabled.

This new pipeline step will prevent that from happening since all the users created in the auth pipeline will have an usable password.

### DIFF
--- a/eox_core/pipeline.py
+++ b/eox_core/pipeline.py
@@ -5,7 +5,7 @@ import logging
 
 from django.db.models.signals import post_save
 
-from eox_core.edxapp_wrapper.users import get_user_profile
+from eox_core.edxapp_wrapper.users import generate_password, get_user_profile
 
 try:
     from social_core.exceptions import AuthFailed, NotAllowedToDisconnect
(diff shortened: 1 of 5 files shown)


eox_core/pipeline.py
```diff
@@ -16,6 +16,23 @@
...
+    properly in the platform if the common.djangoapps.third_party.pipeline.set_logged_in_cookies step is enabled.
+
+    See: https://github.com/eduNEXT/edunext-platform/blob/c83b82a4aba2a5496e3b7da83972a8edf25fcdd2/common/djangoapps/third_party_auth/pipeline.py#L673
+
+    It's recommended to place this step after the social core step that creates the users: (social_core.pipeline.user.create_user).
+    """
+    if user and is_new and not user.has_usable_password():
+        user.set_password(generate_password(length=25))
+        user.save()
+        LOG.info('Assigned an usable password to the user %s on creation.', user)
```
morenol, line 33: I still need here your feedback here to define how we should log in this pipeline step

mariajgrimaldi:
should we add this step in the default pipeline? So all sites use it?

eox_core/pipeline.py
magajh: @morenol I think you could leave this like this and I'll add the calling to the logging_pipeline_step function

magajh (approved):

PR #134, opened 2021-02-04
feat(MTPS): Add more information to pipeline steps

magajh:
This PR adds the logging_pipeline_step function that logs an exception with the information from the failed pipeline step.

### DIFF
--- a/eox_core/logging.py
+++ b/eox_core/logging.py
@@ -1,0 +1,61 @@
+"""
+The logging module defines functions that are used for the logging of the
+pipeline steps.
+"""
+import logging
+import sys
+from pprint import pformat
+
+LOG = logging.getLogger(__name__)
+
+
+# pylint: disable=unused-argument,keyword-arg-before-vararg
+def logging_pipeline_step(level, log_message, **local_vars):
+    """
+    This function logs an info or error level message with the information of the pipeline step.
+    If the logLevel setting is in DEBUG mode then it logs all the arguments passed to the step, if not
+    it will only show the name of the step, user, backend name, redirect_uri and a custom message.
+
+    For example:
+        To set the verbosity level to DEBUG add the setting:
+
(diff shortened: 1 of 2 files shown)


eox_core/logging.py
```diff
@@ -0,0 +1,51 @@
...
+        backend_name=getattr(backend, "name", ""),
+        uri=getattr(backend, "redirect_uri", ""),
+        msg=exception_message,
+    )
+
+    if backend and backend.setting("BACKEND_OPTIONS", {}).get("logLevel", "").lower() == "debug":
+        div = "=" * 100
+        details = local_vars.get("details")
+        args = local_vars.get("args", [])
+        kwargs = local_vars.get("kwargs", {})
```
morenol, line 38: There are a couple of pipeline steps that have as argument: `is_new`, so for these cases it will not be reflected in the kwargs.

Maybe we can remove that from the arguments of these pipeline steps and put at the beginning of each step

is_new = kwargs.get('is_new')

What do you think @magajh ?

magajh: @morenol  Thank you for noticing, I missed that argument. Yes, I think that would be a good solution

morenol (approved):
lgtm

eox_core/pipeline.py
```diff
@@ -80,13 +91,20 @@ def force_user_post_save_callback(auth_entry, is_new, user=None, *args, **kwargs
...
+    is_new = kwargs.get('is_new')
+
     if auth_entry == 'register' and user and is_new:
         user._changed_fields = {'is_active': user.is_active}  # pylint: disable=protected-access
         post_save.send_robust(
             sender=user.__class__,
             instance=user,
             created=False
         )
+        logging_pipeline_step(
```
magajh, line 103: @morenol Just wanted to let you know that I added a log for this step

morenol: It looks fine! @magajh

morenol (approved):

PR #135, opened 2021-02-09
fix: Force the callback of post_save when the user is new

morenol:
This removes the usage of auth_entry in the force_user_post_save_callback pipeline step.

I think that it is safe to remove that and depend on the is_new parameter to decide if we want to send that signal.


magajh (approved):
Looks good

PR #137, opened 2021-03-03
Add a new endpoint to to run celery tasks

MoisesGSalas:
This PR includes a new endpoint that allows to run any arbitrary celery task that is registered in the worker and included in a setting.

### DIFF
--- a/eox_core/api/task_dispatcher/tests/test_views.py
+++ b/eox_core/api/task_dispatcher/tests/test_views.py
@@ -1,0 +1,109 @@
+"""
+Test module for the Task dispatcher API
+"""
+from django.contrib.auth.models import User
+from django.test import TestCase, override_settings
+from django.urls import reverse
+from mock import MagicMock, patch
+from rest_framework import status
+from rest_framework.test import APIClient
+
+from eox_core.api.task_dispatcher.v1.views import TaskAPI
+
+
+@override_settings(EOX_CORE_ASYNC_TASKS=["dummy.task.path.task_name"])
+class TestTaskAPI(TestCase):
+    """ Tests for the task dispatcher API """
+
+    patch_permissions = patch(
+        "rest_framework.permissions.IsAdminUser.has_permission",
+        return_value=True,
+    )
+
+    def setUp(self):
+        """setup fixtures"""
+        self.api_user = User(1, "test@example.com", "test")
+        self.client = APIClient()
+        self.client.force_authenticate(user=self.api_user)
+        self.url = reverse("eox-task-api:task-api")
+
+    @patch("celery.current_app.AsyncResult")
+    @patch_permissions
+    def test_get_task_statuses(self, _, m_async_result):
+        """Test that the GET method works with default parameters"""
+        m_async_result.return_value.status = "READY"
+        m_async_result.return_value.result = "result"
+        m_async_result.return_value.ready.return_value = True
+        expected_response = {"result": "result", "status": "READY"}
+
+        response = self.client.get(self.url, data={"id": "2"})
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(response.data, expected_response)
+
+    @patch.object(TaskAPI, "import_task")
+    @patch("celery.current_app")
+    @patch('eox_core.api.task_dispatcher.v1.views.get_current_site')
+    @patch_permissions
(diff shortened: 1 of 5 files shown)


eox_core/api/task_dispatcher/tests/test_views.py
```diff
@@ -0,0 +1,118 @@
...
+        self.api_user = User(1, "test@example.com", "test")
+        self.client = APIClient()
+        self.client.force_authenticate(user=self.api_user)
+        self.url = reverse("eox-task-api:task-api")
+
+    @patch("celery.current_app.AsyncResult")
+    @patch_permissions
+    def test_get_task_statuses(self, _, m_async_result):
+        """Test that the GET method works with default parameters"""
+
```
mariajgrimaldi, line 34: Can you remove this line?

Try to use this way of grouping inside a test method: Arrange Act Assert 

https://www.c-sharpcorner.com/UploadFile/dacca2/fundamental-of-unit-testing-understand-aaa-in-unit-testing/

eox_core/api/task_dispatcher/v1/views.py
```diff
@@ -0,0 +1,103 @@
...
+        enabled_tasks = settings.EOX_CORE_ASYNC_TASKS
+
+        registered_tasks = celery.current_app.tasks
+
+        if task_path not in enabled_tasks or task_path not in registered_tasks:
+            return Response({"detail": "Invalid task path"}, status=HTTP_404_NOT_FOUND)
+
+        task = self.import_task(task_path)
+
+        result = task.delay(**task_parameters)
```
morenol, line 92: Let's add this + some logging just in case we have a bad configuration in the eox-core settings
```suggestion
        if task:
            result = task.delay(**task_parameters)
```

eox_core/api/task_dispatcher/v1/views.py
```diff
@@ -0,0 +1,103 @@
...
+        return Response({"task_id": result.id})
+
+    @staticmethod
+    def import_task(task_fullpath):
+        """
+        Given the full path to the function in dot notation import
+        the module and return the function reference
+        """
+        module_path, task_name = task_fullpath.rsplit(".", 1)
+        module = import_module(module_path)
```
morenol, line 102: I think that this could raise an exception, we should catch that. Log the error and Return None

morenol (changes requested):

eox_core/api/task_dispatcher/v1/views.py
mariajgrimaldi: You can copy that from this: https://github.com/eduNEXT/eox-hooks/blob/380c426d23b99d65bcbd676d8e20e6534392b6d3/eox_hooks/actions_handler.py#L41

eox_core/api/task_dispatcher/v1/views.py
```diff
@@ -0,0 +1,117 @@
...
+            return Response({"detail": "Invalid task path"}, status=HTTP_404_NOT_FOUND)
+
+        task = self.import_task(task_path)
+
+        if task:
+            result = task.delay(**task_parameters)
+            LOG.info("Task %s with params %s", task_path, task_parameters)
+            return Response({"task_id": result.id})
+
+        return Response({"detail": "Module not found"})
```
morenol, line 100: This shouldn't return a 200 status code

MoisesGSalas: Fixed.

morenol (approved):
lgtm

eox_core/api/task_dispatcher/v1/views.py
```diff
@@ -0,0 +1,117 @@
...
+LOG = logging.getLogger(__name__)
+
+
+class TaskAPI(APIView):
+    """
+    Celery task dispatcher API
+    """
+
+    authentication_classes = [BearerAuthentication]
+    permission_classes = [IsAdminUser]
```
morenol, line 26: I wonder if we could have a use case where we need this for a user with a specific permission for a site.

MoisesGSalas: For the moment I just assumed that the current tasks that could be possibly be run would at least require that level of authorization, and other APIs use those permissions, the eox-core permissions class was obviously discarded because clients have access to those credentials. Maybe in the future we can add a new permission class if needed.

PR #138, opened 2021-03-22
feat: Remove User endpoint.

magajh:
This feature adds an endpoint to safely remove an account from the platform.
This new endpoint is intended to be used in the CLI.

### DIFF
--- a/eox_core/edxapp_wrapper/backends/users_j_v1.py
+++ b/eox_core/edxapp_wrapper/backends/users_j_v1.py
@@ -14,20 +14,29 @@ from openedx.core.djangoapps.lang_pref import LANGUAGE_KEY  # pylint: disable=im
 from openedx.core.djangoapps.site_configuration import helpers as configuration_helpers  # pylint: disable=import-error
 from openedx.core.djangoapps.user_api.accounts import USERNAME_MAX_LENGTH  # pylint: disable=import-error,unused-import
 from openedx.core.djangoapps.user_api.accounts.serializers import UserReadOnlySerializer  # pylint: disable=import-error
+from openedx.core.djangoapps.user_api.accounts.views import \
+    _set_unusable_password  # pylint: disable=import-error,unused-import
+from openedx.core.djangoapps.user_api.models import UserRetirementStatus  # pylint: disable=import-error
 from openedx.core.djangoapps.user_api.preferences import api as preferences_api  # pylint: disable=import-error
 from openedx.core.djangoapps.user_authn.utils import generate_password  # pylint: disable=import-error,unused-import
 from openedx.core.djangoapps.user_authn.views.registration_form import (  # pylint: disable=import-error
     AccountCreationForm,
 )
+from openedx.core.djangolib.oauth2_retirement_utils import \
+    retire_dot_oauth2_models  # pylint: disable=import-error,unused-import
+from rest_framework import status
 from rest_framework.exceptions import NotFound
+from social_django.models import UserSocialAuth  # pylint: disable=import-error
 from student.helpers import create_or_set_user_attribute_created_on_site  # pylint: disable=import-error
-from student.models import (  # pylint: disable=import-error
+from student.models import (  # pylint: disable=import-error,unused-import
     LoginFailures,
+    Registration,
     UserAttribute,
     UserProfile,
     UserSignupSource,
     create_comments_service_user,
     email_exists_or_retired,
+    get_retired_email_by_email,
     username_exists_or_retired,
 )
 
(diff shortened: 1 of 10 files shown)


eox_core/api/v1/views.py
```diff
@@ -200,6 +211,93 @@ def patch(self, request, *args, **kwargs):
         return Response(serialized_user.data)
 
 
+class EdxappRemoveUser(UserQueryMixin, APIView):
```
MoisesGSalas, line 214: This view has quite a bit of business logic attached to it. What if we move that to the users backend? Similar to [`create_edxapp_user`](https://github.com/eduNEXT/eox-core/blob/6080576f28557675e41e1c3964842894063288fc/eox_core/edxapp_wrapper/backends/users_j_v1.py#L63)

eox_core/api/v1/views.py
```diff
@@ -200,6 +211,93 @@ def patch(self, request, *args, **kwargs):
...
+    renderer_classes = (JSONRenderer, BrowsableAPIRenderer)
+
+    def delete(self, request, *args, **kwargs):  # pylint: disable=too-many-locals
+        """
+        Allows to safely remove an edxapp User.
+
+        For example:
+
+        **Requests**:
+            DELETE <domain>/eox-core/api/v1/remove-user/
```
MoisesGSalas, line 230: Why don't we use the existing `/eox-core/api/v1/user/`?

magajh: @MoisesGSalas I was unsure about where to put it. I thought that maybe later we would want to have exclusive permissions for these endpoints that are intended to be used on the CLI, that's why I put it in a different view

MoisesGSalas: In that case what about an endpoint for CLI usage? maybe we can consider this a different API as it's meant to be used by the support team?

magajh: @MoisesGSalas done, I added the support API

eox_core/api/v1/views.py
magajh: done

MoisesGSalas (commented):
Can you add some tests?

magajh:
@MoisesGSalas I didn't add tests for this view since the deletion of the user is made completely in the user's backend by calling different methods of the platform

eox_core/edxapp_wrapper/backends/users_j_v1.py
```diff
@@ -180,6 +190,74 @@ def get_edxapp_user(**kwargs):
...
+                    )
+                    status_code = status.HTTP_200_OK
+
+    msg = "{user} does not have a signup source on the site {site}".format(user=user_response, site=site)
+    status_code = status.HTTP_400_BAD_REQUEST
+    response_dict = {
+        "status": status_code,
+        "message": msg,
+    }
+    return json.dumps(response_dict)
```
MoisesGSalas, line 258: Now that I see this, is there any need for the json dump? wouldn't the `Response` object take care of that?

MoisesGSalas:
@magajh 
> @MoisesGSalas I didn't add tests for this view since the deletion of the user is made completely in the user's backend by calling different methods of the platform.

You can maybe test your permissions at the very least.

eox_core/edxapp_wrapper/backends/users_j_v1.py
magajh: you're right, this was showing the response in a weird format.

magajh:
> @magajh
> 
> > @MoisesGSalas I didn't add tests for this view since the deletion of the user is made completely in the user's backend by calling different methods of the platform.
> 
> You can maybe test your permissions at the very least.

@MoisesGSalas  Done! I added the tests for the permissions

eox_core/edxapp_wrapper/backends/users_j_v1.py
```diff
@@ -180,6 +189,73 @@ def get_edxapp_user(**kwargs):
...
+                Registration.objects.filter(user=user).delete()
+
+                # Delete OAuth tokens associated with the user.
+                retire_dot_oauth2_models(user)
+
+                # Delete user signup source object
+                signup_sources[0].delete()
+
+                msg = "{user} has been removed".format(user=user_response)
+                status_code = status.HTTP_204_NO_CONTENT
```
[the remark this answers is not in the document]
magajh: Done

eox_core/edxapp_wrapper/backends/users_j_v1.py
```diff
@@ -180,6 +189,73 @@ def get_edxapp_user(**kwargs):
...
+                    signup_source.delete()
+
+                    msg = "{user} has more than one signup source. The signup source from the site {site} has been deleted".format(
+                        user=user_response,
+                        site=site,
+                    )
+                    status_code = status.HTTP_200_OK
+    else:
+        msg = "{user} does not have a signup source on the site {site}".format(user=user_response, site=site)
+        status_code = status.HTTP_400_BAD_REQUEST
```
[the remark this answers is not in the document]
magajh: Done

MoisesGSalas (approved):

PR #139, opened 2021-03-22
feat: Replace Username endpoint.

magajh:
This feature adds a new endpoint that handles the replacement of usernames.
Allows updating an Edxapp user's Username along with the forum associated User (comments_service User).

### DIFF
--- a/eox_core/api/support/v1/serializers.py
+++ b/eox_core/api/support/v1/serializers.py
@@ -7,6 +7,17 @@ from __future__ import absolute_import, unicode_literals
 from django.utils import timezone
 from rest_framework import serializers
 
+from eox_core.api.v1.serializers import MAX_SIGNUP_SOURCES_ALLOWED
+from eox_core.edxapp_wrapper.users import (
+    check_edxapp_account_conflicts,
+    get_user_signup_source,
+    get_username_max_length,
+)
+
+UserSignupSource = get_user_signup_source()  # pylint: disable=invalid-name
+
+USERNAME_MAX_LENGTH = get_username_max_length()
+
 
 class WrittableEdxappRemoveUserSerializer(serializers.Serializer):
     """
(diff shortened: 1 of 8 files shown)


eox_core/api/v1/views.py
```diff
@@ -200,6 +203,53 @@ def patch(self, request, *args, **kwargs):
         return Response(serialized_user.data)
 
 
+class EdxappReplaceUsername(UserQueryMixin, APIView):
```
MoisesGSalas, line 206: Same feedback as in #138, maybe use a backend and the existing user endpoint

magajh: Done

eox_core/api/support/v1/serializers.py
```diff
@@ -7,10 +7,59 @@
...
+
+        if self.instance.is_staff or self.instance.is_superuser:
+            raise serializers.ValidationError({"detail": "You can't update users with roles like staff or superuser."})
+
+        if UserSignupSource.objects.filter(user__email=self.instance.email).count() > MAX_SIGNUP_SOURCES_ALLOWED:
+            raise serializers.ValidationError({"detail": "You can't update users with more than one sign up source."})
+
+        return attrs
+
+    def update(self, instance, validated_data):
```
MoisesGSalas, line 56: IIRC, `update` corresponds to `PUT` while `partial_update` is for `PATCH` requests, would you mind checking that?

MoisesGSalas (commented):
can you add a few tests for this?

eox_core/api/support/v1/serializers.py
magajh: Using partial_update raises an error _NotImplementedError: `update()` must be implemented._

MoisesGSalas: You're right. The `partial_update` is for the the `ViewSet` class, I got them mixed up.

magajh:
@MoisesGSalas the tests were added

MoisesGSalas (approved):
Make sure to test everything on stage before merging.

magajh:
> Make sure to test everything on stage before merging.

Ok

PR #140, opened 2021-03-31
docs: adding explanation for the force flag at swagger

felipemontoya:
Adding the explanation for the swagger live documentation.

@MoisesGSalas this parameter is optional, should I mark that somewhere?

Post-Merge

- [ ]  does not require to be released inmediately

### DIFF
--- a/eox_core/api/v1/serializers.py
+++ b/eox_core/api/v1/serializers.py
@@ -186,6 +186,7 @@ class EdxappCourseEnrollmentSerializer(serializers.Serializer):
                     ("username", "johndoe"),
                     ("is_active", True),
                     ("mode", "audit"),
+                    ("force", False),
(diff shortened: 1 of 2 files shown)


MoisesGSalas (approved):
There is no need to mark optional parameters, the approach we are taking is to indicate when a parameter is *required*.

mariajgrimaldi:
can we merge this @MoisesGSalas ?

MoisesGSalas:
> can we merge this @MoisesGSalas ?

@mariajgrimaldi, this could be expanded to include documentation about programs which is currently missing, as well as using the correct serializers for the docs (I think `EdxappCourseEnrollmentQuerySerializer` should be used instead `EdxappCourseEnrollmentSerializer`)

MoisesGSalas:
Closed in favor of #170.

PR #141, opened 2021-04-08
feat: add apiview to get users cohort

mariajgrimaldi:
**Description**

This PR adds a new APIView that uses methods from edx-platform to get the name of a cohort given a course_id and a username. 

**Motivation**

https://3.basecamp.com/3966315/buckets/8050706/todos/3608713852

Before making this PR other alternatives were considered but they didn't fit our use case:

1. [CohortHandler](https://github.com/eduNEXT/edunext-platform/blob/e86e50ffe7b0d596396e27eff3897811d4e63802/openedx/core/djangoapps/course_groups/views.py#L479): 
   GET /api/cohorts/v1/courses/{course_id}/cohorts 
In this case all cohorts will be returned.
2. [CohortUsers](https://github.com/eduNEXT/edunext-platform/blob/e86e50ffe7b0d596396e27eff3897811d4e63802/openedx/core/djangoapps/course_groups/views.py#L574):
   GET /api/cohorts/v1/courses/{course_id}/cohorts/{cohort_id}/users
 In this case we'll get all users associated with a specified cohort.

Perhaps, we could modify CohortUsers to accept the username, and thus check if the user belongs to cohort X.

Donwfalls:

- Although this modification is simple, we would still have to get all cohorts to check one by one if the user belongs to that one.

- Too many requests?

Then this alternative was a better fit.

### DIFF
--- a/eox_core/api/v1/tests/test_cohort.py
+++ b/eox_core/api/v1/tests/test_cohort.py
@@ -1,0 +1,99 @@
+"""
+Test module for the User Cohort API.
+"""
+from django.contrib.auth.models import User
+from django.urls import reverse
+from mock import MagicMock, patch
+from rest_framework import status
+from rest_framework.test import APITestCase
+
+
+class TestUserCohortAPI(APITestCase):
+    """Tests class fort UserCohort API endpoints."""
+
+    def setUp(self):
+        """Setup method for test class."""
+        self.api_user = User(1, "api_user@example.com", "api_user")
+        self.test_user = User(2, "test@example.com", "test")
+        self.client.force_authenticate(user=self.api_user)  # pylint: disable=no-member
+        self.url = reverse("eox-api:eox-api:edxapp-cohort")
+        self.test_user_cohort = MagicMock()
+        self.test_user_cohort.name = "Team 1"
+        self.api_user_cohort = MagicMock()
+        self.api_user_cohort.name = "Team 2"
+
+    @patch("eox_core.api.v1.views.get_user_cohort")
+    @patch("eox_core.api.v1.views.get_valid_course_key")
+    @patch("eox_core.api.v1.views.get_edxapp_user")
+    def test_get_cohort_name(self, get_user_mock, _, get_cohort_mock):
+        """
+        Used to test that the users cohort name can be retrieved given a course and a username.
+        """
+        query_params = {
+            "username": "test",
+            "course_id": "course-v1:edX+DemoX+Demo_Course",
+        }
+        get_user_mock.return_value = self.api_user
+        get_cohort_mock.return_value = self.test_user_cohort
+        expected_response = {
+            "cohort_name": self.test_user_cohort.name,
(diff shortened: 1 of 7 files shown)


eox_core/api/v1/tests/test_cohort.py
```diff
@@ -0,0 +1,101 @@
...
+"""
+from django.contrib.auth.models import User
+from django.test import TestCase
+from django.urls import reverse
+from mock import MagicMock, patch
+from rest_framework import status
+from rest_framework.test import APIClient
+
+
+class TestUserCohortAPI(TestCase):
```
andrey-canon, line 12: https://www.django-rest-framework.org/api-guide/testing/#api-test-cases instead of TestCase

mariajgrimaldi: thanks!

eox_core/api/v1/views.py
```diff
@@ -835,6 +837,73 @@ def _section_breakdown(self, subsection_grades):
...
+        except ValidationError:
+            user = request.user
+
+        course_id = self.query_params.get("course_id")
+
+        if not course_id:
+            raise ValidationError(detail="You have to provide a course_id")
+
+        course_key = get_valid_course_key(course_id)
+        cohort = get_user_cohort(user, course_key, assign=False)
```
MoisesGSalas, line 899: Does it make any difference if you use `use_cached=True`? Can we benefit from that?

mariajgrimaldi: We actually could! Thank you

eox_core/api/v1/views.py
```diff
@@ -835,6 +837,73 @@ def _section_breakdown(self, subsection_grades):
...
+        parameters=[
+            apidocs.query_parameter(
+                name="course_id",
+                param_type=str,
+                description="**required**, Used to identify the course where the cohort must be defined.",
+            ),
+            apidocs.query_parameter(
+                name="username",
+                param_type=str,
+                description="""**optional**, The username used to identify the user that belongs to the cohort.
```
morenol, line 859: It looks like we can query for the cohorts of a user different than me, 

Can this have any security implications that we need to be aware about?

mariajgrimaldi: I can't think of any security implications because we're not sharing critical data. Can you think of any?

I added this to be more flexible but right now I can't think of a specific use case for it, maybe it'd be better if we remove it. We could add it later if needed.

What do you think?

MoisesGSalas: The only thing that comes to mind is that you can confirm a username exists. But how far con you get from there, I'm not so sure. I would say if there isn't any clear use case for now just remove it. If there is any need for that functionality in the future you can add it without breaking backwards compatibility.

mariajgrimaldi:
Proposal moved to xblock: https://github.com/eduNEXT/cohort-api-xblock/pull/1

PR #142, opened 2021-04-29
feat: add TPA pipeline steps used to register signupsources

mariajgrimaldi:
**Description**

This PR adds two pipeline steps used during the TPA process to create signup sources associated with the current user and site.

**Motivation**
https://3.basecamp.com/3966315/buckets/8050706/todos/3695494462

**How to test**

I've been testing the pipeline steps with this custom pipeline using a google provider:

```
    "SOCIAL_AUTH_GOOGLE_OAUTH2_PIPELINE": [
        "third_party_auth.pipeline.parse_query_params",
        "social_core.pipeline.social_auth.social_details",
        "social_core.pipeline.social_auth.social_uid",
        "social_core.pipeline.social_auth.auth_allowed",
        "social_core.pipeline.social_auth.social_user",
        "social_core.pipeline.user.get_username",
        "third_party_auth.pipeline.set_pipeline_timeout",
        "eox_tenant.pipeline.safer_associate_by_email",
        "social_core.pipeline.user.create_user",
        "eox_core.pipeline.ensure_new_user_has_usable_password",
        "social_core.pipeline.social_auth.associate_user",
        "eox_core.pipeline.create_signup_source_for_new_association",  # STEP
        "eox_core.pipeline.register_signup_source",  # STEP
        "social_core.pipeline.social_auth.load_extra_data",
        "social_core.pipeline.user.user_details",
        "eox_core.pipeline.ensure_user_has_profile",
        "eox_core.pipeline.ensure_user_has_signup_source",  # STEP
        "third_party_auth.pipeline.user_details_force_sync",
        "third_party_auth.pipeline.set_id_verification_status",
        "third_party_auth.pipeline.set_logged_in_cookies",
        "third_party_auth.pipeline.login_analytics"
    ],
```

- [x] Tested in stage (after addressing PR comments)

### DIFF
--- a/eox_core/pipeline.py
+++ b/eox_core/pipeline.py
@@ -3,9 +3,15 @@ The pipeline module defines functions that are used in the third party authentic
 """
 import logging
 
+from crum import get_current_request
 from django.db.models.signals import post_save
 
-from eox_core.edxapp_wrapper.users import generate_password, get_user_attribute, get_user_profile
+from eox_core.edxapp_wrapper.users import (
+    generate_password,
+    get_user_attribute,
+    get_user_profile,
+    get_user_signup_source,
+)
 from eox_core.logging import logging_pipeline_step
 
 try:
@@ -14,6 +20,7 @@ except ImportError:
     AuthFailed = object
     NotAllowedToDisconnect = object
 
+UserSignupSource = get_user_signup_source()  # pylint: disable=invalid-name
 LOG = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


andrey-canon (commented):
will those method be implemented together ?
Why do we need two methods, when the code is almost the same ?
I don't see the case when we will implement just one pipeline, so why not just do one ?

mariajgrimaldi:
@andrey-canon mm, I've thought about this, but  @felipemontoya and I came with the idea of implementing two steps for each case to solve.

Some context,  we have two cases:

1. The user has a social auth link but does not have signup source with the site (register_signup_source): executed when social auth links are created
2. The user has neither a social auth link nor signup source with the site (ensure_signup_source): executed when the user is login in

The main reason to have two steps is to use them separately, but I just can't find a real use case. 

If we add just one step it would be something like this:
```
def ensure_user_has_signup_source(user=None, *args, **kwargs):

    if not user or kwargs.get("is_new"):
        return

    _, created = UserSignupSource.objects.get_or_create(user=user, site=get_current_request().site)
    if created:
        logging_pipeline_step(
                "info",
                "Created new singup source for user during the third party pipeline.",
                **locals()
            )
```
And it will be executed in any case when added. This one step will actually solve our current issue and given that I haven't found a use case for the two, I will remove one. (register_signup_source)

Do you agree @felipemontoya?

MoisesGSalas (approved):

felipemontoya:
@mariajgrimaldi thanks a lot.

To answer your question about making this into 2 steps or just one I'd first like to clarify which steps they would be.

ensure_user_has_signup_source -> this is the aggresive one. It means if anyone uses SSO on a site. Even if is the first or 100th time they do, we'll create the signupsource.

create_signup_source_for_new_association -> (this one does not exist yet, and it is the less aggresive) I'm also torn about this one. The logic for this is that the mere action of "login in" should not necessarily create a record assigning the User to the site. I believe we would want to use the "ensure_" for our service but I can see customer that do not rely on signupsource wanting to use it differently.

eox_core/pipeline.py
```diff
@@ -164,3 +171,24 @@ def assert_user_information(details, user, backend, *args, **kwargs):
...
+def ensure_user_has_signup_source(user=None, *args, **kwargs):
+    """
+    This pipeline step registers a new signup source for the current user and site.
+
+    It's recommended to place this step at the end of the TPA pipeline, e.g., after the step
+    'eox_core.pipeline.ensure_user_has_profile',
+
+    To avoid errors, these steps avoid signup source creation during the registration process.
+    """
+    if not user or kwargs.get("is_new"):
```
felipemontoya, line 185: Could you please explain in the docstring or a comment where is this is_new comming from?

I don't really care about skipping this for is_new. I am however making the push to have 2 steps. One that filters by ["new_association"](https://github.com/python-social-auth/social-core/blob/master/social_core/pipeline/social_auth.py#L54) and one that does not.

felipemontoya: Also on the topic of `kwargs.get("is_new")`. Should someone put this step before [social_user](https://github.com/python-social-auth/social-core/blob/master/social_core/pipeline/social_auth.py#L28) we would have a key_error rendering a very confusing 500 error to the final user.

Would you consider using `kwargs.get("is_new", False)`?

felipemontoya (changes requested):

eox_core/pipeline.py
mariajgrimaldi: Thanks for your suggestions, I updated the PR taking them into account.

eox_core/pipeline.py
```diff
@@ -164,3 +171,51 @@ def assert_user_information(details, user, backend, *args, **kwargs):
...
+    It's recommended to place this step after the creation of the social auth link, i.e, after the step
+    'social_core.pipeline.social_auth.associate_user',
+
+    To avoid errors or unnecessary searches, these steps are skipped using "is_new" key insterted by
+    'social_core.pipeline.user.create_user' after creating a user.
+    """
+    if not user or kwargs.get("is_new", False):
+        return
+
+    if kwargs.get("new_association"):
```
felipemontoya, line 190: Same applies here. kwargs.get("new_association", False):

felipemontoya (approved):
I left a minor fix that needs to be applied. Otherwise I'm ready to merge

PR #143, opened 2021-05-06
feat: updated backends to fit lilac release

mariajgrimaldi:
In lilac `from student... import...` is deprecated in favor of `common.djangoapps.student.models`


eox_core/settings/common.py
```diff
@@ -20,9 +20,9 @@ def plugin_settings(settings):
     See: https://github.com/edx/edx-platform/blob/master/openedx/core/djangoapps/plugins/README.rst
     """
     settings.EOX_CORE_COMMENTS_SERVICE_USERS_BACKEND = "eox_core.edxapp_wrapper.backends.comments_service_users_j_v1"
-    settings.EOX_CORE_USERS_BACKEND = "eox_core.edxapp_wrapper.backends.users_j_v1"
+    settings.EOX_CORE_USERS_BACKEND = "eox_core.edxapp_wrapper.backends.users_l_v1"
```
morenol, line 23: I think that we should not change the defaults here yet. At least we should wait for the official release of lilac

mariajgrimaldi: But the DOD is plugins working on lilac. Do you think there will be a big change till the release?

morenol: After these changes is the plugin working fine with Lilac? I think that there won'tbe a big change till the release, but in any I think that it is safer to keep the defaults without  Lilac. We are still using the main branch of eox-core in the SaaS and this could break that.

mariajgrimaldi: Yes! The only issue I found is how to import the student module. Alright,  I'll follow your advice and let everyone know we are still using juniper backends as the default. Thank you!

morenol (approved):

eox_core/edxapp_wrapper/backends/pre_enrollment_l_v1.py
```diff
@@ -0,0 +1,117 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+"""
+Backend for the pre-enrollment (white listings) functionality that works under the open-release/hawthorn.beta1 tag
+"""
+# pylint: disable=import-error, protected-access
+from __future__ import absolute_import, unicode_literals
```
andrey-canon, line 7: I think this is not required in python 3

eox_core/edxapp_wrapper/backends/pre_enrollment_l_v1.py
```diff
@@ -0,0 +1,117 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+"""
+Backend for the pre-enrollment (white listings) functionality that works under the open-release/hawthorn.beta1 tag
```
magajh, line 4: hawthorn?

eox_core/edxapp_wrapper/backends/users_l_v1.py
```diff
@@ -0,0 +1,323 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+"""
+Backend for the create_edxapp_user that works under the open-release/juniper.master tag
```
magajh, line 4: juniper

eox_core/edxapp_wrapper/backends/pre_enrollment_l_v1.py
mariajgrimaldi: From what I read you're right! I'll remove it

PR #145, opened 2021-05-11
Fixed relative import statements

shaidar:
Minor fixes to relative imports that are causing migration failures during the ansible run.

### DIFF
--- a/eox_core/edxapp_wrapper/backends/enrollment_h_v1.py
+++ b/eox_core/edxapp_wrapper/backends/enrollment_h_v1.py
@@ -17,7 +17,7 @@ from openedx.core.djangoapps.content.course_overviews.models import CourseOvervi
 from openedx.core.lib.exceptions import CourseNotFoundError
 from pytz import utc
 from rest_framework.exceptions import APIException, NotFound
-from student.models import CourseEnrollment
+from common.djangoapps.student.models import CourseEnrollment
 
 from eox_core.edxapp_wrapper.backends.edxfuture_i_v1 import get_program
 from eox_core.edxapp_wrapper.coursekey import get_valid_course_key, validate_org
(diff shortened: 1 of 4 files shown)


felipemontoya:
Thanks @shaidar, we will review this shortly.

@andrey-canon and @mariajgrimaldi I tagged you both to see if you can review this. Would it create errors when we run this in older installations (hawthorn based for example)?

mariajgrimaldi:
Hey there @felipemontoya and @shaidar! 

We encountered these issues while making eox-core forward compatible with lilac, to solve them we decided to create new backends for [user](https://github.com/eduNEXT/eox-core/blob/master/eox_core/edxapp_wrapper/backends/users_l_v1.py) and [pre-enrollment](https://github.com/eduNEXT/eox-core/blob/master/eox_core/edxapp_wrapper/backends/pre_enrollment_l_v1.py) and leave the current (working in juniper) defaults in [settings/common.py](https://github.com/eduNEXT/eox-core/blob/fa484526ab5cfeb160fd19f055a8d40a0444a6ea/eox_core/settings/common.py#L23).

So, what I recommend to avoid errors in other installations is to override settings/common.py with the correct backends for lilac.

@shaidar can you test changing EOX_CORE_USERS_BACKEND and EOX_CORE_PRE_ENROLLMENT_BACKEND?

For example:

EOX_CORE_USERS_BACKEND: "eox_core.edxapp_wrapper.backends.users_l_v1"
EOX_CORE_PRE_ENROLLMENT_BACKEND: "eox_core.edxapp_wrapper.backendspre_enrollment_l_v1"

shaidar:
@mariajgrimaldi Great. Thanks for the info. Will give that a shot and let you know.

mariajgrimaldi:
I added a new PR for the missing backend for enrollments #146. I'll let you know when it's approved!

shaidar:
@mariajgrimaldi Looks like we need a new tag as the `4.8` one doesn't include the `users_l_v1` module. I can pull from master, but just thought I'd mention it.

mariajgrimaldi:
Of course, I'll merge the enrollment backend and release the new tag. Thank you! @shaidar

mariajgrimaldi:
The new tag is out! @shaidar :smile:

mariajgrimaldi:
Master already supports lilac.

PR #149, opened 2021-05-26
docs: update documentation with newer info (after juniper)

mariajgrimaldi:
Concerns: what can we remove from the read me that might be deprecated? or what can we add that's missing


MoisesGSalas (approved):

andrey-canon (commented):
You could include the eox-audit-model option

andrey-canon (approved):

PR #153, opened 2021-06-23
feat: add override extra_data method to OIDC

mariajgrimaldi:
### Description
This PR adds an override method to the OIDC custom backend. The method adds useful information to the extra_data field from the UserSocialAuth model.

### How to test
1. OIDC configuration
2. There's a pipeline step called `load_extra_data` that sets the new info every time the OIDC pipeline is executed (eg. when login in). This pipeline step belongs to the `DEFAULT_AUTH_PIPELINE`. So, log in to the site using the OIDC provider

Check extra_data and you'll get:

More context: https://3.basecamp.com/3966315/buckets/8050706/todos/3834823891

### DIFF
--- a/eox_core/social_tpa_backends.py
+++ b/eox_core/social_tpa_backends.py
@@ -51,3 +51,13 @@ class ConfigurableOpenIdConnectAuth(OpenIdConnectAuth):
                 self.OIDC_ENDPOINT + '/.well-known/openid-configuration'
             )
         return self.configuration
+
+    def extra_data(self, user, uid, response, details=None, *args, **kwargs):  # pylint: disable=keyword-arg-before-vararg
+        """
+        Override method extra_data returned by the provider including user
+        information.
+        """
+        data = super().extra_data(user, uid, response, details, *args, **kwargs)
+        data["user_details"] = self.get_user_details(response)
+
+        return data


eox_core/social_tpa_backends.py
```diff
@@ -51,3 +51,13 @@ def oidc_config(self):
                 self.OIDC_ENDPOINT + '/.well-known/openid-configuration'
             )
         return self.configuration
+
+    def extra_data(self, user, uid, response, details=None, *args, **kwargs):  # pylint: disable=keyword-arg-before-vararg
+        """
+        Override method extra_data returned by the provider including user
+        information.
+        """
+        data = super(ConfigurableOpenIdConnectAuth, self).extra_data(user, uid, response, details, *args, **kwargs)
```
MoisesGSalas, line 60:
```suggestion
        data = super().extra_data(user, uid, response, details, *args, **kwargs)
```
Since we dropped support for Python 2.

MoisesGSalas (changes requested):

felipemontoya (approved):
I'd say let's change the key to 'user_details', but otherwise this looks good to me.

MoisesGSalas (approved):

PR #154, opened 2021-06-24
feat(TPA): override get_user_id to prepend a slug to the uid.

MoisesGSalas:
# Description
In a multi tenancy context uniquely identifying social associations by `user_id` and provider becomes insufficient. This PR attempts to alleviate the current situation by prepending a slug included in the provider configuration. This way each OIDC configuration would have their own namespace.

To ease the deployment of the feature, previous associations that lack the slug would be updated at login time. Once all the existing associations have been updated, is possible to forbid any new association that doesn't include an slug.
The whole feature can be enabled on a per site basis.

### DIFF
--- a/eox_core/social_tpa_backends.py
+++ b/eox_core/social_tpa_backends.py
@@ -11,7 +11,10 @@ try:
     from social_core.backends.open_id_connect import OpenIdConnectAuth
 except ImportError:
     OpenIdConnectAuth = object
-
+try:
+    from social_core.exceptions import AuthMissingParameter
+except ImportError:
+    AuthMissingParameter = Exception
 
 configuration_helper = get_configuration_helper()  # pylint: disable=invalid-name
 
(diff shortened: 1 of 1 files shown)


eox_core/social_tpa_backends.py
```diff
@@ -9,7 +9,9 @@
 
 try:
     from social_core.backends.open_id_connect import OpenIdConnectAuth
+    from social_core.exceptions import AuthMissingParameter
```
felipemontoya, line 12: Do you think we need to split this into two try blocks, if the auth missing fails but OpenId works then we would override it, but that is very unlikely to happen, right?

eox_core/social_tpa_backends.py
```diff
@@ -61,3 +63,44 @@ def extra_data(self, user, uid, response, details=None, *args, **kwargs):  # pyl
...
+        in the 'OTHER_SETTINGS' field from the provider configuration.
+        In case a slug was not defined raise an exception due to misconfigured
+        backend.
+        This three behaviours can be controlled via the settings:
+            - 'SOCIAL_AUTH_NAMESPACED_UIDS': if false behave as originally
+            intended. If true, add a slug to the uid.
+            - 'SOCIAL_AUTH_ALLOW_SLUGLESS_UID': if false raise and exception when no slug
+            is defined. If true, return de original uid when no slug is defined.
+            - 'SOCIAL_AUTH_ALLOW_WRITE_SLUG_UID': If true update the uid of existing social
+            users if any.
```
felipemontoya, line 80: Could you please add an example here of what the settings json should have for this to:1) convert old style into new and 2) work only when the new format is updated in all the socialauths

eox_core/social_tpa_backends.py
```diff
@@ -61,3 +63,44 @@ def extra_data(self, user, uid, response, details=None, *args, **kwargs):  # pyl
...
+        slug for social users that don't have one. The slug must be defined
+        in the 'OTHER_SETTINGS' field from the provider configuration.
+        In case a slug was not defined raise an exception due to misconfigured
+        backend.
+        This three behaviours can be controlled via the settings:
+            - 'SOCIAL_AUTH_NAMESPACED_UIDS': if false behave as originally
+            intended. If true, add a slug to the uid.
+            - 'SOCIAL_AUTH_ALLOW_SLUGLESS_UID': if false raise and exception when no slug
+            is defined. If true, return de original uid when no slug is defined.
+            - 'SOCIAL_AUTH_ALLOW_WRITE_SLUG_UID': If true update the uid of existing social
```
felipemontoya, line 79: I would suggest that this is always true and that we never want to "not update". As long as the "slug" field is created we want to update.

eox_core/social_tpa_backends.py
```diff
@@ -61,3 +63,44 @@ def extra_data(self, user, uid, response, details=None, *args, **kwargs):  # pyl
...
+        strategy = self.strategy
+        slug = strategy.setting('slug', backend=self)
+        provider = self.name
+        allow_slugless_uid = self.setting('ALLOW_SLUGLESS_UID', False)
+        allow_write_slug_uid = self.setting('ALLOW_WRITE_SLUG_UID', True)
+
+        if not slug:
+            if allow_slugless_uid:
+                return uid
+            raise AuthMissingParameter(self, 'slug')
```
felipemontoya, line 96: This is what happens when I dont put anything in the settings right?

Does this mean the change breaks stuff and we need to bump major? I'm not convinced we need to, but I dont know how do we mark this in the code, the repo or the Changelog.

Alternatively, we make it so not adding anything to settings makes it behave exactly as it does now.

felipemontoya (changes requested):

eox_core/social_tpa_backends.py
MoisesGSalas: It's very unlikely, but covering that 1/1000 case doesn't require too much effort.

eox_core/social_tpa_backends.py
MoisesGSalas: Sure thing.

eox_core/social_tpa_backends.py
MoisesGSalas: The idea was that once all the old entries have been updated we can avoid a few queries to de DB since all the new entries will always have a slug if 'slug' is defined. But if you think it only adds complexity we can go with always true.

eox_core/social_tpa_backends.py
MoisesGSalas: This will only happen if you set `'SOCIAL_AUTH_NAMESPACED_UIDS'` to `true` but don't define a slug in `OTHER_SETTINGS`. It won't break anything unless you are making a conscious effort to use this feature but don't configure it correctly.

andrey-canon (approved):
LGTM

eox_core/social_tpa_backends.py
felipemontoya: Then it is ok

felipemontoya (approved):

felipemontoya:
Thanks @MoisesGSalas I think everything I asked is covered now

PR #155, opened 2021-06-29
refactor: add debug mode when loading extra_data

mariajgrimaldi:
This PR adds debug mode when loading extra_data during the Auth pipeline execution. It also wraps get_user_details in try-except which allows saving information even when failing.

### DIFF
--- a/eox_core/social_tpa_backends.py
+++ b/eox_core/social_tpa_backends.py
@@ -57,11 +57,55 @@ class ConfigurableOpenIdConnectAuth(OpenIdConnectAuth):
 
     def extra_data(self, user, uid, response, details=None, *args, **kwargs):  # pylint: disable=keyword-arg-before-vararg
         """
-        Override method extra_data returned by the provider including user
-        information.
+        Get information returned by the provider including user information.
+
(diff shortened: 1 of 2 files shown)


andrey-canon (approved):
you could test this behavior

eox_core/social_tpa_backends.py
```diff
@@ -57,11 +57,55 @@ def oidc_config(self):
...
         data = super().extra_data(user, uid, response, details, *args, **kwargs)
-        data["user_details"] = self.get_user_details(response)
+
+        debug_mode_enabled = self.setting('DEBUG_ENABLED', False)
+
+        try:
+            data["user_details"] = self.get_user_details(response)
+            if debug_mode_enabled:
+                data["userinfo_response"] = response
+        except Exception:  # pylint: disable=broad-except
```
MoisesGSalas, line 106: The only thing that comes to mind is that we were going to do some work to the oidc exceptions in the future, maybe re-rasing would be better? I think is better to revisit this if needed. Let's go with this. LGTM

MoisesGSalas (approved):

felipemontoya (approved):

PR #159, opened 2021-07-01
feat: Allow adding profile fields when updating a user

magajh:
This PR allows updating the UserProfile fields considered as 'safe'
These "safe" fields are defined in the setting  EOX_CORE_USER_UPDATE_SAFE_FIELDS

A modification was also made to the update user endpoint to allow filtering the user by _username_ or _email_

### DIFF
--- a/eox_core/api/v1/serializers.py
+++ b/eox_core/api/v1/serializers.py
@@ -20,6 +20,7 @@ from eox_core.edxapp_wrapper.users import (
     get_username_max_length,
 )
 from eox_core.utils import (
+    create_user_profile,
     get_gender_choices,
     get_level_of_education_choices,
     get_registration_extra_fields,
@@ -171,7 +172,7 @@ class EdxappExtendedUserSerializer(EdxappUserSerializer):
                     self.fields[field_name] = serializers.CharField(**serializer_field)
 
 
-class WrittableEdxappUserSerializer(EdxappUserSerializer):
+class WrittableEdxappUserSerializer(EdxappExtendedUserSerializer):
     """
     Handles the serialization of the user data required to update an edxapp user.
     """
@@ -206,14 +207,50 @@ class WrittableEdxappUserSerializer(EdxappUserSerializer):
         """
         Update method for safe fields.
         """
+        extended_profile_fields = getattr(settings, 'extended_profile_fields', [])
+        # Obtain only the User profile fields defined in the EdxappExtendedUserSerializer
+        all_fields = set(EdxappExtendedUserSerializer().fields)
+        non_profile_fields = set(EdxappUserSerializer().fields)
+        extra_registration_fields = all_fields - non_profile_fields
+        # Check if the User has a Profile
+        has_profile = hasattr(instance, 'profile')
+
+        extra_registration_fields.update(extended_profile_fields)
+        if has_profile:
+            profile_meta = instance.profile.get_meta()
+
         for key, value in validated_data.items():
 
             if key == "password":
                 instance.set_password(value)
+
+            elif key in extra_registration_fields or key == "fullname":
+                if not has_profile:
+                    create_user_profile(instance)
+                    profile_meta = {}
+                    has_profile = True
+
+                # First check if the field belongs to the meta
+                if key in extended_profile_fields:
+                    profile_meta[key] = value
+
+                # Key is one of the user profile fields
+                else:
+                    if key == "year_of_birth":
+                        value = int(value)
+                    # 'name' field is sent as 'fullname' in the request
+                    if key == "fullname":
+                        key = "name"
+                    setattr(instance.profile, key, value)
+
             else:
                 setattr(instance, key, value)
 
         if validated_data:
+            if has_profile:
+                # Update user profile meta
+                instance.profile.set_meta(profile_meta)
+                instance.profile.save()
             instance.save()
 
         return instance
--- a/eox_core/api/v1/tests/test_users.py
+++ b/eox_core/api/v1/tests/test_users.py
@@ -4,13 +4,96 @@ Test module for users viewset.
 from django.contrib.auth.models import User
 from django.test import TestCase, override_settings
 from django.urls import reverse
-from mock import patch
+from mock import MagicMock, patch
 from rest_framework import status
+from rest_framework.serializers import ValidationError
 from rest_framework.test import APIClient
 
-from eox_core.api.v1.serializers import EdxappUserQuerySerializer
+from eox_core.api.v1.serializers import EdxappUserQuerySerializer, WrittableEdxappUserSerializer
+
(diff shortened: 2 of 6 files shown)


eox_core/api/v1/serializers.py
```diff
@@ -196,14 +198,47 @@ def update(self, instance, validated_data):
         """
         Update method for safe fields.
         """
+        # Obtain only the User profile fields defined in the EdxappExtendedUserSerializer
+        all_fields = set(EdxappExtendedUserSerializer().fields)
+        non_profile_fields = set(EdxappUserSerializer().fields)
+        extra_registration_fields = all_fields - non_profile_fields
+        # Check if the User has a Profile
+        has_profile = hasattr(instance, 'profile')
```
mariajgrimaldi, line 206: can you use a default?

eox_core/api/v1/serializers.py
```diff
@@ -196,14 +198,47 @@ def update(self, instance, validated_data):
...
+
+        extended_profile_fields = getattr(settings, 'extended_profile_fields', [])
+        extra_registration_fields.update(extended_profile_fields)
+        if has_profile:
+            profile_meta = instance.profile.get_meta()
```
mariajgrimaldi, line 211: why do you get profile_meta here and in line 220 you empty it?

magajh: @mariajgrimaldi thank you for noticing, I forgot to add the condition `if not has_profile` to ensure that this only happens if the user currently does not have a UserProfile and a profile field wants to be added

eox_core/api/v1/tests/test_users.py
```diff
@@ -78,68 +160,216 @@ def test_update_role_not_allowed(self, get_edxapp_user, signup_source, _):
...
-        "name": "personal_id",
-        "type": "text",
-    },
-    {
-        "label": "Mobile",
-        "name": "mobile",
-        "type": "text",
-        "restrictions": {
-            "max_length": 12,
+        self.assertEqual(status.HTTP_200_OK, response.status_code)
```
mariajgrimaldi, line 181: you can also check the update (?)

eox_core/api/v1/tests/test_users.py
```diff
@@ -78,68 +160,216 @@ def test_update_role_not_allowed(self, get_edxapp_user, signup_source, _):
...
+            "country": "ve",
+        }
+        user_serializer.return_value.data = {}
+        get_edxapp_user.return_value = self.user
+        signup_source.objects.filter.return_value.count.return_value = 1
+
+        serializer = WrittableEdxappUserSerializer(self.user, data=update_data, partial=True)
+        serializer.is_valid()
+
+        self.assertEqual(validated_data, serializer.data)
```
mariajgrimaldi, line 262: how are you checking that can't be updated?

magajh: @mariajgrimaldi  Basically what I'm testing here is that the WrittableEdxappUserSerializer updates its fields according to the site settings. The validated data that is being returned by the serializer only has the fields that are properly configured in the extra registration fields settings, if a field that is not there is sent in the request data, the serializer simply ignores it.
Maybe I should change the name of the test

eox_core/api/v1/tests/test_users.py
magajh: @mariajgrimaldi what do you mean? 
I don't think it's really possible to check the update since the UserProfile is a model defined in the edx-platform. Here I'm just checking the request does not return any Validation error when trying to update the fields

mariajgrimaldi (approved):

eox_core/api/v1/tests/test_users.py
```diff
@@ -4,13 +4,94 @@
...
+    "goals": "required",
+    "mailing_address": "hidden",
+    "mobile": "required",
+    "mother_last_name": "required",
+    "personal_id": "required",
+    "type_document": "required",
+}
+
+
+@override_settings(extended_profile_fields=EXTENDED_PROFILE_FIELDS)
```
MoisesGSalas, line 78: Im pretty sure you can user `@override_settings` once and pass al the settings as params

magajh: Done!

eox_core/settings/common.py
```diff
@@ -46,7 +46,8 @@ def plugin_settings(settings):
     settings.EOX_CORE_USER_ORIGIN_SITE_SOURCES = ['fetch_from_unfiltered_table', ]
     settings.EOX_CORE_APPEND_LMS_MIDDLEWARE_CLASSES = False
     settings.EOX_CORE_ENABLE_UPDATE_USERS = True
-    settings.EOX_CORE_USER_UPDATE_SAFE_FIELDS = ["is_active", "password", "fullname"]
+    settings.EOX_CORE_USER_UPDATE_SAFE_FIELDS = ["is_active", "password", "fullname", "mailing_address", "year_of_birth", "gender", "level_of_education", "city", "country", "goals", "bio", "phone_number"]
+    settings.EOX_CORE_USER_UPDATE_NON_SAFE_META_FIELDS = []
```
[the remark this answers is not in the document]
magajh: nice catch

eox_core/utils.py
```diff
@@ -102,3 +102,11 @@ def set_select_custom_field(custom_field, serializer_field):
             raise serializers.ValidationError({"{}".format(field_name): "The default value must be one of the options"})
 
     return serializer_field
+
+
+def check_user_profile(user):
```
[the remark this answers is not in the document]
magajh: Done

eox_core/api/v1/serializers.py
```diff
@@ -189,9 +190,10 @@ def validate(self, attrs):
         """
         # If at least one of these conditions is true, then the user can't be updated.
         safe_fields = getattr(settings, "EOX_CORE_USER_UPDATE_SAFE_FIELDS", [])
+        unsafe_meta_fields = getattr(settings, 'EOX_CORE_USER_UPDATE_UNSAFE_META_FIELDS', [])
 
         for attr in attrs:
-            if attr not in safe_fields:
+            if attr not in safe_fields or attr in unsafe_meta_fields:
```
andrey-canon, line 196: I don't understand why we need this if we have `EOX_CORE_USER_UPDATE_SAFE_FIELDS = ['email']` this won't raise any exception but if we have `EOX_CORE_USER_UPDATE_SAFE_FIELDS = []` this will raise the exception so why do we need to do `EOX_CORE_USER_UPDATE_UNSAFE_META_FIELDS= ['email']` if the result is the same with `EOX_CORE_USER_UPDATE_SAFE_FIELDS = []`

magajh: You are right, the setting is unnecessary

PR #162, opened 2021-07-15
feat: add post_register through API signal

mariajgrimaldi:
This PR sends a signal after the user registration through `POST /eox-core/api/v1/user/`.  The signal is placed in a similar location as when registering through the LMS `/register` page, here the reference: https://github.com/eduNEXT/edunext-platform/pull/510

### DIFF
--- a/eox_core/edxapp_wrapper/backends/users_j_v1.py
+++ b/eox_core/edxapp_wrapper/backends/users_j_v1.py
@@ -25,6 +25,7 @@ from openedx.core.djangoapps.user_authn.views.registration_form import (  # pyli
 )
 from openedx.core.djangolib.oauth2_retirement_utils import \
     retire_dot_oauth2_models  # pylint: disable=import-error,unused-import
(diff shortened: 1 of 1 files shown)


magajh (approved):

MoisesGSalas (approved):

eox_core/edxapp_wrapper/backends/users_j_v1.py
```diff
@@ -155,6 +157,9 @@ def create_edxapp_user(*args, **kwargs):
     except Exception:  # pylint: disable=broad-except
         errors.append("No comments_service_user was created")
 
+    # Announce registration through API
```
magajh, line 160: @mariajgrimaldi  The signal will be sent If the [transaction](https://github.com/eduNEXT/eox-core/blob/master/eox_core/edxapp_wrapper/backends/users_j_v1.py#L133) fails. Did you test what happens in this case?

magajh (approved):

andrey-canon (approved):

PR #165, opened 2021-07-19
feat: send signal after user registration through API

mariajgrimaldi:
PR #162 was closed in favor of this one.

This PR sends a signal after the user registration through `POST /eox-core/api/v1/user/`.  The signal is placed in a similar location as when registering through the LMS `/register` page, here is the reference: https://github.com/eduNEXT/edunext-platform/pull/510. It uses a backend to import the signal.

Context: https://3.basecamp.com/3966315/buckets/8050706/todos/3933592446

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,6 +11,7 @@ Change Log
 .. There should always be an "Unreleased" section for changes pending release.
 Unreleased
 ----------
+Send post_register signal after user registration through API call.
 
 [4.13.2] - 2021-07-16
 ---------------------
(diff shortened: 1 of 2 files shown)


magajh:
@mariajgrimaldi Do you think it's necessary to create a backend for the signals instead of importing them in the backend file of the related object? Either way seems fine to me

magajh (approved):

mariajgrimaldi:
@magajh at this point, I'd rather go with the easiest (a try-except like you said)

MoisesGSalas (approved):

PR #168, opened 2021-08-06
fix: wrap openedx imports in try-except block to avoid import errors

mariajgrimaldi:
**Description**

This PR wraps course and coursekey backends in a try-except block to avoid import errors when executing tests in other packages

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,6 +11,9 @@ Change Log
 .. There should always be an "Unreleased" section for changes pending release.
 Unreleased
 ----------
+Changed
+~~~~~~~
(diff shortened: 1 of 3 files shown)


MoisesGSalas:
@mariajgrimaldi just for curiosity, what packages were broken that we didn't realize until now?

mariajgrimaldi:
> @mariajgrimaldi just for curiosity, what packages were broken that we didn't realize until now?

I added eox-core as a dependency in eox-hooks, and this is happening:



That was the solution -so we didn't have to create a new backend in eox-hooks-. I'm  also doing it to use the courses backend

MoisesGSalas (approved):

PR #170, opened 2021-08-10
docs: add enrollment force doc and change serializer for response doc

MaferMazu:
Documentation changes:

1. I added information about the enrollment force flag
2. Add force to EdxappCourseEnrollmentQuerySerializer
3. Use EdxappCourseEnrollmentQuerySerializer in `post` docs 

### About force
I got the information about the `force` by comparing the behavior of the two functions `_force_create_enrollment `and `_create_or_update_enrollment` in `eox-core/eox_core/edxapp_wrapper/backends/enrollment_h_v1.py`. This last function raises several exceptions such as:
EnrollmentClosedError
CourseFullError
AlreadyEnrolledError
and it also verifies that the enrollment mode belongs to the modes allowed by the course, while the force function skips them.

### DIFF
--- a/eox_core/api/v1/views.py
+++ b/eox_core/api/v1/views.py
@@ -534,9 +534,9 @@ class EdxappEnrollment(UserQueryMixin, APIView):
     renderer_classes = (JSONRenderer, BrowsableAPIRenderer)
 
     @apidocs.schema(
-        body=EdxappCourseEnrollmentSerializer,
+        body=EdxappCourseEnrollmentQuerySerializer,
         responses={
-            200: EdxappCourseEnrollmentSerializer,
+            200: EdxappCourseEnrollmentQuerySerializer,
             202: "User doesn't belong to site.",
             400: "Bad request, invalid course_id or missing either email or username.",
         },
@@ -554,6 +554,7 @@ class EdxappEnrollment(UserQueryMixin, APIView):
               "username": "johndoe",
               "course_id": "course-v1:edX+DemoX+Demo_Course",
               "mode": "audit",
+              "force": "False",
               "is_active": "False",
               "enrollment_attributes": [
                 {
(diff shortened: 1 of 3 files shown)


eox_core/api/v1/views.py
```diff
@@ -581,6 +582,11 @@ def post(self, request, *args, **kwargs):
         - `is_active` (boolean, _body_):
             Flag indicating whether the enrollment is active.
 
+        - `force` (boolean, _body_):
+            Flag indicating whether the platform business rules for enrollment must be skipped. When it is true, the enrollment
+            is created without looking at the enrollment dates, if the course is full, if the enrollment mode is part of the modes
```
mariajgrimaldi, line 587: are you sure the course mode is not checked? https://github.com/eduNEXT/eox-core/blob/a6d976a6cec34a64cbc028912030f8c5fc34aa03/eox_core/edxapp_wrapper/backends/enrollment_h_v1.py#L207

MaferMazu: yes, this function checks that the mode belongs to the defined modes, but if it is forced it skips the check if a mode belongs to a course.
https://github.com/eduNEXT/eox-core/blob/a6d976a6cec34a64cbc028912030f8c5fc34aa03/eox_core/edxapp_wrapper/backends/enrollment_h_v1.py#L316-L318

eox_core/api/v1/views.py
```diff
@@ -710,7 +716,7 @@ def put(self, request, *args, **kwargs):
             ),
         ],
         responses={
-            200: EdxappCourseEnrollmentSerializer,
+            200: EdxappCourseEnrollmentQuerySerializer,
```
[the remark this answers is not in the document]
MaferMazu: You're right

eox_core/api/v1/views.py
mariajgrimaldi: what about this line? https://github.com/eduNEXT/eox-core/blob/a6d976a6cec34a64cbc028912030f8c5fc34aa03/eox_core/edxapp_wrapper/backends/enrollment_h_v1.py#L311

andrey-canon (approved):

andrey-canon (approved):

PR #174, opened 2021-08-31
Exposing the option to set the deployment environment for Sentry

blarghmatey:
Sentry allows for setting the `environment` attribute to identify in which stage of deployment an error event occurs. This can help with determining why an error happened and the urgency required to diagnose and resolve an issue.


mariajgrimaldi:
FYI @jfavellar90 @NeOneSoft @felipemontoya

felipemontoya:
Thanks @blarghmatey, this looks like an oversight from our side since we are used to change this via our eox-tenant plugin.
I think this looks good, but before merging I'd like someone from the ops/sre team to look at it.

:+1:

jfavellar90:
Hi @blarghmatey IGTM :+1: . I think when the environment value is set to `None`, the setting does not take effect in the integration, can you confirm this is correct?

PR #175, opened 2021-09-24
fix: truncate OIDC first_name to field max length

mariajgrimaldi:
**Description**
This PR truncates the user's first_name sent by the identity provider to avoid exceeding the Django limit.

**Before**
Error 500:
image

**After**
Registers the user and:
image

The screenshots were taken after testing on stage.

### DIFF
--- a/eox_core/social_tpa_backends.py
+++ b/eox_core/social_tpa_backends.py
@@ -4,6 +4,7 @@ Extensions to the regular defined third party auth backends
 import logging
 
 from django.conf import settings
+from django.contrib.auth import get_user_model
 
 from eox_core.edxapp_wrapper.configuration_helpers import get_configuration_helper
 
@@ -19,6 +20,7 @@ except ImportError:
 configuration_helper = get_configuration_helper()  # pylint: disable=invalid-name
 
 LOG = logging.getLogger(__name__)
+User = get_user_model()  # pylint: disable=invalid-name
 
 
 class ConfigurableOpenIdConnectAuth(OpenIdConnectAuth):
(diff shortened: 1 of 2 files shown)


eox_core/social_tpa_backends.py
```diff
@@ -4,6 +4,7 @@
 import logging
 
 from django.conf import settings
+from django.contrib.auth.models import User
```
MoisesGSalas, line 7:
```suggestion
from django.contrib.auth import get_user_model
```
And the corresponding `User = get_user_model()`

MoisesGSalas (changes requested):

eox_core/social_tpa_backends.py
mariajgrimaldi: ufff right. Thanks

eox_core/social_tpa_backends.py
```diff
@@ -55,6 +57,19 @@ def oidc_config(self):
...
 
+    def get_user_details(self, response):
+        """
+        Override used to truncate first name in case is longer than the field's
+        length.
+        """
+        user_details = super().get_user_details(response)
+        first_name_max_len = User._meta.get_field("first_name").max_length  # pylint: disable=no-member, protected-access
+
+        if len(user_details.get("first_name", "")) > first_name_max_len:
```
MoisesGSalas, line 68: You don't even have to check length, if the stop value of the slice is greater than the original length it returns the whole thing.

MoisesGSalas (approved):

magajh (approved):

jignaciopm (approved):

MaferMazu (approved):

PR #184, opened 2021-11-08
Run JS quality test on a different build.

MoisesGSalas:
## Description

Separate the Javascript quality tests from the main run so it doesn't block the PyPi release. Fix the readme formatting.

## Checklist for Merge

- [x] Tested in a remote environment / NA
- [x] Updated documentation / NA
- [x] Rebased master/main
- [x] Squashed commits

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,6 +21,10 @@ workflows:
                 debian_version: "stretch"
               - python_version: "3.5"
                 debian_version: "buster"
+      - quality:
+          filters:
+            tags:
+              only: /v?[0-9]+(\.[0-9]+)*/
       - pypi:
           requires:
             - test
@@ -66,7 +70,6 @@ jobs:
             test -d venv || virtualenv venv
             . venv/bin/activate
             pip install -r requirements/tox.txt
-            npm install
 
       # run tests!
       # this example uses Django's built-in test-runner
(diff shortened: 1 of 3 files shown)


jfavellar90:
The reason why we are using the image `circleci/python:3.5-node` to run javascript tests is that it includes the Node version `12.18.4`. I think Node 12 was the base to build the Course Manager react app. Should we include this fact somewhere in the docs? or do something like [this](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#engines) to specify the node version the project works on? or definitively move this React app to its own repository?

In order to move forward, I give thumbs up to this PR, we can solve this issue in a future iteration.

.circleci/config.yml
```diff
@@ -21,6 +21,10 @@ workflows:
                 debian_version: "stretch"
               - python_version: "3.5"
                 debian_version: "buster"
+      - quality:
```
jfavellar90, line 24: @MoisesGSalas what do you think about `quality-js` as job name to differentiate it from the python jobs?

MoisesGSalas: That would be a better name yes.

MoisesGSalas:
I used `circleci/python:3.5-node` because the image was already there. The test should also pass on node `v14.18.1` (that's the version I have in my local environment). I could run the tests on both machines (3.5 and 3.8) if is worth it to know they are failing on node 16, maybe use a specific node image for that. As long as they don't block new releases.

I'm partial to the last option, I think the React app should be moved elsewhere, but there should be a discussion about that first. But for the time being I think those test shouldn't block new releases,

jfavellar90 (approved):

jfavellar90:
@MoisesGSalas I'll proceed with the merge. Feel free to move the React app discussion to the right place so we can get feedback from the team and take action.

PR #208, opened 2022-05-29
fix: get settings course_org_filter when validating org for async proc

mariajgrimaldi:
## Description

[get_current_site_orgs()](https://github.com/openedx/edx-platform/blob/open-release/lilac.master/openedx/core/djangoapps/site_configuration/helpers.py#L216) obtains the current site orgs getting first the current site:

image
image
image
image

 which doesn't exist in async processes. So, this adds an alternative: if the current org is not found then try getting them directly from the site settings.

## Testing instructions

I'm currently testing this with an async task that enrolls users into a course. It's in a private repository, reach out if you want to test it out.

## Additional information

I found this issue while trying to enroll users in a course using an async task and configuring the site with course_org_filter. With this change, it's working correctly.

## Checklist for Merge

- [x] Updated documentation
- [x] Rebased master/main
- [x] Squashed commits

### DIFF
--- a/eox_core/edxapp_wrapper/backends/coursekey_m_v1.py
+++ b/eox_core/edxapp_wrapper/backends/coursekey_m_v1.py
@@ -1,0 +1,53 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+"""
+Backend for the CourseKey validations that works under the open-release/maple.master tag
+"""
+# pylint: disable=import-error, protected-access
+from __future__ import absolute_import, unicode_literals
+
+from django.conf import settings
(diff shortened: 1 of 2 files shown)


mariajgrimaldi:
Question: should we fix the current backend or add a new one for maple?

Alec4r:
> Question: should we fix the current backend or add a new one for maple?

I prefer to create a new backend file because we have a hawthorn backend.

mariajgrimaldi:
Can you review again please? @Alec4r

Alec4r (approved):

JuanDavidBuitrago (approved):

PR #222, opened 2022-11-01
feat: Make extra registration fields optional in during edxapp account creation

magajh:
## Description

In this PR we make some changes to the [create_edxapp_user()](https://github.com/eduNEXT/eox-core/blob/master/eox_core/edxapp_wrapper/backends/users_m_v1.py#L104) in order to make the extra registration fields optional when registering a new user un edxapp.

Three conditions are checked in order to be able to skip these extra registration fields:
    1. The skip_extra_registration_fields field is sent in the data for the account creation as True.
    2. The user making the request is staff.
    3. The data received in the parameters does not contain any of the REGISTRATION_EXTRA_FIELDS configured for the microsite.

In case any of these conditions are not met, the function will request the extra registration fields to be sent for the creation of the new user.

## Why are these changes necessary?

 In PR #223 we are implementing an endpoint that creates a new OAuth Application. 
We need to create a user in the platform to set it as the owner of the new edxapp application and want to use the [create_edxapp_user()](https://github.com/eduNEXT/eox-core/blob/master/eox_core/edxapp_wrapper/backends/users_m_v1.py#L104) function for this. However, this function requires to send the extra registration fields configured for the site during the account creation, so we are making these changes to be able to register a user only with the basic information such as fullname, username and email, without having to worry about other registration fields.

## Testing instructions

The best way to test this change is through the endpoint that creates an Oauth application, implemented in PR #223, since we are not sending any of the extra registration fields during the creation of the user that will be the owner of the application.
On the contrary, if you try to send only these fields
```
{
    "username": "mariecurie",
    "email": "mariecurie@example.com",
    "fullname": "Marie Curie",
}
```
In the endpoint that handles the [creation of a new user](https://openedx.edunext.co/eox-core/api/v1/user/),  you will get form errors for not sending the extra registration fields.

## Checklist for Merge

- [ ] Tested in a remote environment
- [ ] Updated documentation
- [x] Rebased master/main
- [x] Squashed commits

### DIFF
--- a/eox_core/edxapp_wrapper/backends/users_j_v1.py
+++ b/eox_core/edxapp_wrapper/backends/users_j_v1.py
@@ -3,11 +3,12 @@
 """
 Backend for the create_edxapp_user that works under the open-release/juniper.master tag
 """
-# pylint: disable=import-error
 from __future__ import absolute_import, unicode_literals
 
 import logging
 
+# pylint: disable=import-error
+from crum import get_current_user
 from django import forms
 from django.conf import settings
 from django.contrib.auth import get_user_model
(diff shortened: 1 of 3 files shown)


Alec4r:
@magajh Maybe we need new unit tests or update the unit tests? if not, This looks good for me.

magajh:
@Alec4r  I initially thought about updating the tests, but then I checked and there aren't any tests for the backend modules

Alec4r (approved):

JuanDavidBuitrago (dismissed):
It's working as expected!!

dcoa (dismissed):

JuanDavidBuitrago (approved):

Alec4r (approved):

dcoa (approved):

PR #239, opened 2022-12-26
chore(DS-354): update requirements

Henrrypg:
## Description

This PR es to update requirements, specificly openedx-events due to https://edunext.atlassian.net/browse/DS-354 fix


JuanDavidBuitrago (dismissed):

Makefile
```diff
@@ -39,7 +39,7 @@ clean: ## Remove generated byte code, coverage reports, and build artifacts
 python-test: clean ## Run tests
 	$(TOX) pip install -r requirements/test.txt --exists-action w
 	$(TOX) coverage run --source="." -m pytest ./eox_core
-	$(TOX) coverage report --fail-under=70
+	$(TOX) coverage report --fail-under=60
```
MaferMazu, line 42: Could you please tell me why you needed to change this?

MaferMazu: Maybe we are analyzing the covering of something we don't need to cover.

JuanDavidBuitrago (approved):

Makefile
Henrrypg: Thank you @MaferMazu, you're rigth, we dont need to change this line, the commit is updated, can you please review it?

requirements/base.txt
```diff
@@ -1,16 +1,16 @@
 #
-# This file is autogenerated by pip-compile with python 3.10
-# To update, run:
+# This file is autogenerated by pip-compile with Python 3.8
```
Alec4r, line 2: Why was it changed? Why are we using 3.8 instead of 3.10?

Henrrypg: That's my bad @Alec4r, by default i use python3.8, i just updated it again with python.310, thanks

Alec4r (approved):

JuanDavidBuitrago (approved):

PR #248, opened 2023-03-02
feat: allow sentry to receive extra params

Ian2012:
## Description

This PR fixes the issue #247 

## Testing instructions

Set the variable `EOX_CORE_SENTRY_EXTRA_OPTIONS` to any custom param that sentry accepts

### DIFF
--- a/eox_core/settings/production.py
+++ b/eox_core/settings/production.py
@@ -138,15 +138,24 @@ def plugin_settings(settings):  # pylint: disable=function-redefined
 
     if sentry_sdk is not None and sentry_integration_dsn is not None:
         from eox_core.integrations.sentry import ExceptionFilterSentry  # pylint: disable=import-outside-toplevel
-        sentry_sdk.init(
-            before_send=ExceptionFilterSentry(),
-            dsn=sentry_integration_dsn,
-            environment=sentry_environment,
-            integrations=[
-                DjangoIntegration(),
-            ],
-            # If you wish to associate users to errors (assuming you are using
-            # django.contrib.auth) you may enable sending PII data.
-            send_default_pii=True,
-            **sentry_extra_options
-        )
+        try:
+            sentry_sdk.init(
+                before_send=ExceptionFilterSentry(),
+                dsn=sentry_integration_dsn,
+                environment=sentry_environment,
+                integrations=[
+                    DjangoIntegration(),
+                ],
+                send_default_pii=True,
+                **sentry_extra_options
+            )
+        except TypeError:
+            sentry_sdk.init(
+                before_send=ExceptionFilterSentry(),
+                dsn=sentry_integration_dsn,
+                environment=sentry_environment,
+                integrations=[
+                    DjangoIntegration(),
+                ],
+                send_default_pii=True
+            )
--- a/eox_core/settings/production.py
+++ b/eox_core/settings/production.py
@@ -131,6 +131,10 @@ def plugin_settings(settings):  # pylint: disable=function-redefined
         'EOX_CORE_SENTRY_ENVIRONMENT',
         settings.EOX_CORE_SENTRY_ENVIRONMENT
     )
+    sentry_extra_options = getattr(settings, 'ENV_TOKENS', {}).get(
+        'EOX_CORE_SENTRY_EXTRA_OPTIONS',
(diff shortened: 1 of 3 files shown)


mariajgrimaldi:
@Ian2012: can you give us an example of what to add to `EOX_CORE_SENTRY_EXTRA_OPTIONS`

eox_core/settings/production.py
```diff
@@ -141,7 +145,7 @@ def plugin_settings(settings):  # pylint: disable=function-redefined
             integrations=[
                 DjangoIntegration(),
             ],
-
+            **sentry_extra_options,
```
mariajgrimaldi, line 148: can we move this to line 151?

mariajgrimaldi: what if `sentry_extra_options` is invalid?

mariajgrimaldi:
Could you document these changes [here](https://github.com/eduNEXT/eox-core/#integrations-with-third-party-services)? Please

eox_core/settings/production.py
Ian2012: What would be an invalid `sentry_extra_option`? What cases should I cover?

Ian2012:
> @DonatoBD @DeimerM @Ian2012: this is a minor change but I'd like to test it remotely to check its behavior. Although, I don't know if creating a sentry project just for a test would be worth it. Have you tested this approach before?

@mariajgrimaldi We have not tested it, but we can do it in our stage environment

mariajgrimaldi:
@Ian2012: please, do. Let us know when it's deployed so we can test it too. Also, it would be awesome if you could attach the testing instructions using sentry. Thanks!

eox_core/settings/production.py
mariajgrimaldi: An option that's not handled by sentry_sdk.init, or does the method just ignores the options that don't recognize?

Ian2012:
@mariajgrimaldi We performed a test using other variables and it works as expected.
```python
debug=True
```
I get a different error when trying to configure profiling:

EOX_CORE_SENTRY_EXTRA_OPTIONS = {
    "_experiments": { "profiles_sample_rate": 1.0, },
    "traces_sample_rate": 1.0
}
```
[sentry] DEBUG: [Profiling] Discarding profile because transaction is discarded.
--- Logging error ---
Traceback (most recent call last):
  File "/opt/pyenv/versions/3.8.12/lib/python3.8/logging/__init__.py", line 440, in format
    return self._format(record)
  File "/opt/pyenv/versions/3.8.12/lib/python3.8/logging/__init__.py", line 436, in _format
    return self._fmt % record.__dict__
KeyError: 'userid'
```

It seems it's a problem with the default logging format of edx-platform.

BTW, this functionality works and the above issue shouldn't be a blocker

eox_core/settings/production.py
```diff
@@ -141,8 +145,8 @@ def plugin_settings(settings):  # pylint: disable=function-redefined
             integrations=[
                 DjangoIntegration(),
             ],
-
             # If you wish to associate users to errors (assuming you are using
             # django.contrib.auth) you may enable sending PII data.
-            send_default_pii=True
+            send_default_pii=True,
+            **sentry_extra_options
```
mariajgrimaldi, line 151: If I add a duplicated option (I did it by accident), instances won't be able to initialize. So can you please add a try-except for this?
image

mariajgrimaldi: Instances won't initialize either (at least workers) if the option doesn't exist:
image

mariajgrimaldi (changes requested):

eox_core/settings/production.py
Ian2012: I've added a try-except block that will ignore the `sentry_extra_options` param if sentry fails to initialize

Ian2012: What do you think? @mariajgrimaldi

Ian2012: Just a a reminder @mariajgrimaldi

mariajgrimaldi (approved):
Since this has backward compatibility if a new incompatible variable is configured then the sentry configuration will default to the old configuration (without extra option compatibilities). So this is technically safe to merge!

PR #251, opened 2023-09-05
feat: add string represantion to exc_value

johanv26:
# Description

## Problem
Some hint in exception dont have `log_record` so exc_text is `""`

## Solution

Log record info now for sentry seems to be for breadcrumbs. Eox core Sentry config is using `before_send` and not `before_breadcrumbs` so some exceptions are managed without the log_record. https://docs.sentry.io/platforms/python/guides/django/configuration/filtering/hints/ So the self.exc_text is `''`.
If this is empty regexes don't make anything in
https://github.com/eduNEXT/eox-core/blob/v6.1.0/eox_core/integrations/sentry.py#L154 so this change add `exc_value` if nothing is in log_records, the idea is to could compare at least with the str of the `exc_value`.
Peek 2023-08-30 16-37

## Before 
Regex doesn't work for exceptions with higher uses like Django exceptions.
Eg. When a Django model uses a get and I want to ignore it, I want to ignore only one particular model and not all the Django model object Does not exist. So the idea is that regex would work.
``` json
"EOX_CORE_SENTRY_IGNORED_ERRORS": [
      {
        "exc_class": "django.core.exceptions.ObjectDoesNotExist",
        "exc_text": [".*EnterpriseCustomerIdentityProvider matching query does not exist.*"]
      }
]
```
## After

Peek 2023-08-30 16-33

**extra info** 
This was deployed to prod in nelp version that the only difference is this change to avoid sending the enterprise error.
https://github.com/eduNEXT/eox-core/tree/v6.1.0-nelp
https://github.com/eduNEXT/eox-core/pull/250

## Checklist for Merge

- [x] Updated documentation
- [ ] Rebased master/main
- [ ] Squashed commits

### DIFF
--- a/eox_core/integrations/sentry.py
+++ b/eox_core/integrations/sentry.py
@@ -76,6 +76,8 @@ class ExceptionFilterSentry:
         if 'exc_info' in hint:
             _exc_type, exc_value, _tb = hint['exc_info']
             self.exc_value = exc_value
+            if not self.exc_text:
+                self.exc_text = str(exc_value)
 
     def clear_exception_values(self):
         """


MaferMazu (approved):
This change looks good to me; it is a good improvement. And we can have it for our next release.

dcoa (approved):

PR #252, opened 2023-11-06
refactor!: remove openedx-events constraint to use what's installed

mariajgrimaldi:
## Description

This PR removes the current constraint on `openedx-events` so instead of re-installing that specified version uses the existent one. This avoids forcing a version on  newer Open edX versions that use openedx-events > 0.13.0

## Testing instructions

This is what happens when installing openedx-events in a quince development environment:
image

Now, after the change, it uses the version available:
image

## Additional information

Open edX plugins usually use a [setup.py](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#setup-py) file for configurations where it specifies which requirements the library needs when installed. These requirements can specify a version to be installed, which we currently do with openedx-events. Now, in a plugin context we should use the library version tha's currently installed (if installed), since openedx-events is already part of the base requirements of Open edX, we should not change the version.

## Checklist for Merge

- [ ] Tested in a remote environment
- [ ] Updated documentation
- [ ] Rebased master/main
- [ ] Squashed commits


mariajgrimaldi:
The Python 3.10, Django 3.2 suite is currently failing due to the libraries update: https://github.com/eduNEXT/eox-core/actions/runs/6774203929/job/18410767530, so I'm gonna drop the upgrade of the packages not related to openedx-events for now. But so you know that after upgrading using `make upgrade` the CI is failing.

MaferMazu (approved):
Thanks for this, Majo.

PR #254, opened 2023-11-17
Mfmz/palm support

MaferMazu:
## Description

Describe what this pull request changes, and why. Include implications for people using this change.

Useful information to include:
- Which user roles will this change impact? Common user roles are "Learner", "Course Author",
"Developer", and "Operator".
- Include screenshots for changes to the UI (ideally, both "before" and "after" screenshots, if applicable).
- Provide links to the description of corresponding configuration changes. Remember to correctly annotate these
changes.

## Testing instructions

Please provide detailed step-by-step instructions for testing this change.

## Additional information

Include anything else that will help reviewers and consumers understand the change.
- Does this change depend on other changes elsewhere?
- Any special concerns or limitations? For example: deprecations, migrations, security, or accessibility.
- Link to other information about the change, such as Jira issues, GitHub issues, or Discourse discussions.

## Checklist for Merge

- [ ] Tested in a remote environment
- [ ] Updated documentation
- [ ] Rebased master/main
- [ ] Squashed commits

PR #259, opened 2023-11-24
feat: added logs to errors imports

BetoFandino:
Logs in errors imports

## Description
This change is due to the difficulty in tracking errors in some imports. They will help to locate faster any errors or warnings that may occur that may hinder testing and debugging. 

Useful information to include:
- Which user roles will this change impact: Developer.

## Checklist for Merge

- [ x] Tested in a remote environment / NA
- [x ] Updated documentation / NA
- [ x] Rebased master/main / NA
- [x ] Squashed commits / NA


eox_core/settings/production.py
```diff
@@ -1,13 +1,18 @@
...
 from .common import *  # pylint: disable=wildcard-import, unused-wildcard-import
 
+LOG = logging.getLogger(__name__)
+
 try:
     import sentry_sdk
     from sentry_sdk.integrations.django import DjangoIntegration
 except ImportError:
     sentry_sdk = DjangoIntegration = None
+    LOG.error("ImportError while importing %s", ImportError)
```
mariajgrimaldi, line 15: what does this log? don't you need to include the exception instead of `ImportError`?

MaferMazu:
Thanks for this @BetoFandino.
In #254, we already added `social-auth-core` as a requirement for this plugin to avoid the try-except errors related to `social_core`. But we probably still need your proposal for other packages, for example, sentry. Let me know when you require a review for this 🙌

BetoFandino:
Hi @MaferMazu, super thank you very much but then I ask you two little things: 
     - Can we assume that they are going to apply that for sentry and therefore we close this pr and wait for you? or do you need our help to push that?
     - I'm not sure if you removed it but that PR you put, if you made social-core mandatory then it would be good to remove this too https://github.com/eduNEXT/eox-core/blob/master/setup.py#L76 ?

eox_core/settings/production.py
BetoFandino: Hi @mariajgrimaldi  sorry, yes I think if you need to include the exception. But it depends on the answer if you are going to do the same for sentry as what was done in PR [#254](https://github.com/eduNEXT/eox-core/pull/254), then it would no longer be necessary I think.

MaferMazu:
You already have this PR; we can continue the process of adding the try-except to sentry imports in your PR, and in the palm support PR, handle the` social_core` imports. About the TPA requirements, thanks for that, I didn't notice, I will remove it in #254

BetoFandino:
Ok @MaferMazu, then I will make the pertinent changes in this PR. Thank you very much!!! 😃

MaferMazu:
@BetoFandino, do you still want to merge this?

BetoFandino:
@MaferMazu if possible yes please

MaferMazu:
@BetoFandino, can you fix the issues, please?

bra-i-am:
hi! I hope you're well. 

@BetoFandino, just a friendly reminder about this PR so we can continue the merging process, thank you c:

MaferMazu:
Thanks for updating this, @BetoFandino. I only have a question: Why do we need the file coursekey_h_v1? In my understanding, that file is to support the Open edX Release Hawthorn, and right now, we have a coursekey_m_v1 that supports Maple and all the next releases.

The rest looks good to me. When we figure out the Hawthorn backend, we will be good to merge.

MaferMazu (approved):
Looks good to me

dcoa (approved):

PR #265, opened 2024-05-07
feat: add user lang cookie preference

dcoa:
This PR has the intention to allow users to change the website language preference based on their cockies.

This PR is inspired by https://github.com/eduNEXT/eox-nelp/pull/127 and allows the normal behavior of the header and footer lang selectors for Django templates.

https://edunext.atlassian.net/browse/DS-936


MaferMazu (approved):
Thanks for this. It's working as expected. Before this change, I tried to manually change the openedx-lang cookies, but the platform didn't take them into account; with this change, the platform takes into account the preferences I set in the cookies.
Thanks @dcoa 🙌

image
For future reference, I will let the initial problem here: https://github.com/openedx/edx-platform/blob/open-release/palm.master/openedx/core/djangoapps/lang_pref/middleware.py#L61-L62.

bra-i-am (approved):
Working as expected!

Before:
image

After:
image

PR #266, opened 2024-05-08
fix: add basename to router register to avoid conflicts with services

mariajgrimaldi:
## Description

This PR solves this issue raised in one of our installations after building and deploying an image with this package installed:

Screenshot from 2024-05-07 12-31-44

It looks like [this API endpoint](https://github.com/eduNEXT/eox-core/blob/b1bc5954e2a0936476f1fcd54dfb2d0edcc17e1b/eox_core/api/data/v1/routers.py#L20-L21) conflicts with the enrollments API from the edx-platform. I'm not sure why, though because our installation has been working fine for some time.


BryanttV:
Hi @mariajgrimaldi, we were trying to replicate the error, but it was not possible. However, we will keep the error on our radar. Thanks!

BryanttV:
Hi @mariajgrimaldi, in [this PR](https://github.com/eduNEXT/eox-core/pull/273) we add the basename in all the routes. There should no longer be any errors since `v10.5.0`

PR #280, opened 2024-09-19
feat: add integration tests for the enrollments api `AP-1346`

BryanttV:
## Description
This PR adds integration tests for the enrollment API. These tests are executed in the job of **Tutor Integration Tests**

This PR is part of a set of PRs to add integration tests to the enrollment API. It must be separated into different PRs because doing it in one PR exceeds the maximum size allowed by the **pr-size-labeler** action.

- https://github.com/eduNEXT/eox-core/pull/281
- https://github.com/eduNEXT/eox-core/pull/282

## Testing instructions

## Jira Issue
- [AP-1346](https://edunext.atlassian.net/browse/AP-1346)

[AP-1346]: https://edunext.atlassian.net/browse/AP-1346?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


eox_core/api/v1/tests/integration/test_views.py
```diff
@@ -370,6 +431,595 @@ def test_update_user_in_tenant_user_not_found(self, param: str, value: str) -> N
...
+
+        Open edX definitions tested:
+        - `get_edxapp_user`
+        - `create_enrollment`
+        - `check_edxapp_account_conflicts`
+        - `check_edxapp_enrollment_is_valid`
+
+        Expected result:
+        - The status code is 200.
+        - The enrollment is created successfully in the tenant with the provided data.
```
mariajgrimaldi, line 457: For this to be true, we'd have to check for the new enrollment in the database, which is more or less the `test_get_enrollment_success` test case. So, I'd modify this to be: 

```suggestion
        - The response indicates the enrollment was created successfully. 
```

eox_core/api/v1/tests/integration/test_views.py
```diff
@@ -370,6 +431,595 @@ def test_update_user_in_tenant_user_not_found(self, param: str, value: str) -> N
...
+        self.assertEqual(response_data["mode"], enrollment_data["mode"])
+        self.assertEqual(response_data["course_id"], enrollment_data["course_id"])
+        self.assertTrue(response_data["is_active"])
+        self.assertIn("created", response_data)
+
+    @ddt.data(
+        ("mode", {"mode": ["This field is required."]}),
+        ("course_id", {"non_field_errors": ["You have to provide a course_id or bundle_id"]}),
+        ("username", {"non_field_errors": ["Email or username needed"]}),
+    )
```
mariajgrimaldi, line 481: Should we also consider when neither the username nor the email are passed?

eox_core/api/v1/tests/integration/test_views.py
```diff
@@ -370,6 +431,595 @@ def test_update_user_in_tenant_user_not_found(self, param: str, value: str) -> N
...
+        response = self.create_enrollment(self.tenant_x, enrollment_data)
+
+        response_data = response.json()
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertEqual(response_data, error)
+
+    @ddt.data("email", "username")
+    def test_force_create_enrollment_success(self, param: str) -> None:
+        """
+        Test force creating a enrollment with a valid user, course and mode in a tenant.
```
mariajgrimaldi, line 512:
```suggestion
        Test force creating an enrollment with a valid user, course and mode in a tenant.
```

eox_core/api/v1/tests/integration/test_views.py
```diff
@@ -370,6 +431,595 @@ def test_update_user_in_tenant_user_not_found(self, param: str, value: str) -> N
...
+        enrollment_data.pop(param)
+
+        response = self.create_enrollment(self.tenant_x, enrollment_data)
+
+        response_data = response.json()
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertEqual(response_data, error)
+
+    @ddt.data("email", "username")
+    def test_force_create_enrollment_success(self, param: str) -> None:
```
mariajgrimaldi, line 510: What's the difference between forcing the enrollment and not that we're testing here?

eox_core/api/v1/tests/integration/test_views.py
```diff
@@ -370,6 +431,595 @@ def test_update_user_in_tenant_user_not_found(self, param: str, value: str) -> N
...
+        response_data = response.json()
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(response_data["username"], user_data["username"])
+        self.assertEqual(response_data["mode"], enrollment_data["mode"])
+        self.assertEqual(response_data["course_id"], enrollment_data["course_id"])
+        self.assertTrue(response_data["is_active"])
+        self.assertIn("created", response_data)
+
+    @ddt.data("email", "username")
+    def test_create_valid_course_mode_invalid_user(self, param: str) -> None:
```
mariajgrimaldi, line 544: Where's the non-existent username or email here? Can we make it more explicit?

eox_core/api/v1/tests/integration/test_views.py
```diff
@@ -370,6 +431,595 @@ def test_update_user_in_tenant_user_not_found(self, param: str, value: str) -> N
...
+        - `create_enrollment`
+        - `check_edxapp_account_conflicts`
+        - `check_edxapp_enrollment_is_valid`
+
+        Expected result:
+        - The status code is 200.
+        - The enrollment is created successfully in the tenant with the provided data.
+        """
+        user_data = next(FAKE_USER_DATA)
+        self.create_user(self.tenant_x, user_data)
```
mariajgrimaldi, line 525: Do we have to create the users each time? Can we make it part of the setup?

eox_core/api/v1/tests/integration/test_views.py
```diff
@@ -370,6 +431,595 @@ def test_update_user_in_tenant_user_not_found(self, param: str, value: str) -> N
...
+        """
+        Test creating a enrollment with a valid course, valid mode, and a non-existent user in a tenant.
+
+        Open edX definitions tested:
+        - `check_edxapp_enrollment_is_valid`
+
+        Expected result:
+        - The status code is 400.
+        - The response contains an error message about the user not found.
+        - The enrollment is not created in the tenant.
```
mariajgrimaldi, line 554: Item 7 of the [integration tests guidance document](https://internal.docs.edunext.co/en/latest/internal/Teams/Descontrol/Internal_processes/integration_tests_in_plugins/integration_tests_with_openedx_components.html?highlight=integration%20tests#pruebas-de-apis) says we should also check the application's state after each update. I don't know if we did that for the users API, though, not for every test. Do you think we should do that here, or did we decide not to check explicitly because of performance issues?

eox_core/api/v1/tests/integration/test_views.py
```diff
@@ -370,6 +431,595 @@ def test_update_user_in_tenant_user_not_found(self, param: str, value: str) -> N
...
+        Open edX definitions tested:
+        - `check_edxapp_enrollment_is_valid`
+
+        Expected result:
+        - The status code is 202.
+        - The response contains an error message about the user not found on the tenant.
+        - The enrollment is not created in the tenant.
+        """
+        user_data = next(FAKE_USER_DATA)
+        self.create_user(self.tenant_y, user_data)
```
mariajgrimaldi, line 583: ^ same comment here about creating the users in the setup method instead.

eox_core/api/v1/tests/integration/test_views.py
```diff
@@ -370,6 +431,595 @@ def test_update_user_in_tenant_user_not_found(self, param: str, value: str) -> N
...
+
+        response_data = response.json()
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertIn("course_id", response_data)
+        self.assertEqual(response_data["course_id"], [f"Invalid course_id {self.course_id}"])
+
+    @ddt.data("email", "username")
+    def test_create_valid_user_course_invalid_mode(self, param: str) -> None:
+        """
+        Test creating a enrollment with a valid user, valid course, and a not available mode in a tenant.
```
mariajgrimaldi, line 658:
```suggestion
        Test creating an enrollment with a valid user, valid course, and an unavailable mode in a tenant.
```

eox_core/api/v1/tests/integration/test_views.py
```diff
@@ -370,6 +431,595 @@ def test_update_user_in_tenant_user_not_found(self, param: str, value: str) -> N
...
+
+        response_data = response.json()
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertIn("non_field_errors", response_data)
+        self.assertEqual(response_data["non_field_errors"], ["Course not found"])
+
+    @ddt.data("email", "username")
+    def test_create_valid_user_mode_invalid_course_for_tenant(self, param: str) -> None:
+        """
+        Test creating a enrollment with a valid user, valid mode, and a course from another tenant.
```
mariajgrimaldi, line 630:
```suggestion
        Test creating an enrollment with a valid user, valid mode, and a course from another tenant.
```

eox_core/api/v1/tests/integration/test_views.py
```diff
@@ -370,6 +431,595 @@ def test_update_user_in_tenant_user_not_found(self, param: str, value: str) -> N
...
+        self.assertEqual(response.status_code, status.HTTP_202_ACCEPTED)
+        self.assertEqual(
+            response_data["error"]["detail"],
+            f"No user found by {{'{param}': '{enrollment_data[param]}'}} on site {self.tenant_x['domain']}.",
+        )
+
+    @ddt.data("email", "username")
+    def test_create_valid_user_mode_invalid_course(self, param: str) -> None:
+        """
+        Test creating a enrollment with a valid user, valid mode, and a non-existent course in a tenant.
```
mariajgrimaldi, line 602:
```suggestion
        Test creating an enrollment with a valid user, valid mode, and a non-existent course in a tenant.
```

eox_core/api/v1/tests/integration/test_views.py
```diff
@@ -370,6 +431,595 @@ def test_update_user_in_tenant_user_not_found(self, param: str, value: str) -> N
...
+
+        response_data = response.json()
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertIn("non_field_errors", response_data)
+        self.assertEqual(response_data["non_field_errors"], ["User not found"])
+
+    @ddt.data("email", "username")
+    def test_create_valid_course_mode_invalid_user_for_tenant(self, param: str) -> None:
+        """
+        Test creating a enrollment with a valid course, valid mode, and a user from another tenant.
```
mariajgrimaldi, line 572:
```suggestion
        Test creating an enrollment with a valid course, valid mode, and a user from another tenant.
```

eox_core/api/v1/tests/integration/test_views.py
```diff
@@ -370,6 +431,595 @@ def test_update_user_in_tenant_user_not_found(self, param: str, value: str) -> N
...
+        self.assertEqual(response_data["username"], user_data["username"])
+        self.assertEqual(response_data["mode"], enrollment_data["mode"])
+        self.assertEqual(response_data["course_id"], enrollment_data["course_id"])
+        self.assertTrue(response_data["is_active"])
+        self.assertIn("created", response_data)
+
+    @ddt.data("email", "username")
+    def test_create_valid_course_mode_invalid_user(self, param: str) -> None:
+        """
+        Test creating a enrollment with a valid course, valid mode, and a non-existent user in a tenant.
```
mariajgrimaldi, line 546:
```suggestion
        Test creating an enrollment with a valid course, valid mode, and a non-existent user in a tenant.
```

eox_core/api/v1/tests/integration/test_views.py
```diff
@@ -370,6 +431,595 @@ def test_update_user_in_tenant_user_not_found(self, param: str, value: str) -> N
...
+
+    @ddt.data(
+        ("mode", {"mode": ["This field is required."]}),
+        ("course_id", {"non_field_errors": ["You have to provide a course_id or bundle_id"]}),
+        ("username", {"non_field_errors": ["Email or username needed"]}),
+    )
+    @ddt.unpack
+    def test_create_enrollment_missing_required_fields(self, param: str, error: list | dict) -> None:
+        """
+        Test creating a enrollment with missing required fields.
```
mariajgrimaldi, line 485:
```suggestion
        Test creating an enrollment with missing required fields.
```

eox_core/api/v1/tests/integration/test_views.py
```diff
@@ -370,6 +431,595 @@ def test_update_user_in_tenant_user_not_found(self, param: str, value: str) -> N
...
+
+        response_data = response.json()
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertIn("non_field_errors", response_data)
+        self.assertEqual(response_data["non_field_errors"], ["Mode not found"])
+
+    @ddt.data("email", "username")
+    def test_force_create_valid_user_course_mode_not_allowed(self, param: str) -> None:
+        """
+        Test creating a enrollment with a valid user, valid course, and a not available mode in a tenant.
```
mariajgrimaldi, line 686:
```suggestion
        Test creating an enrollment with a valid user, valid course, and an unavailable mode in a tenant.
```

mariajgrimaldi: What's the difference between this test case and `test_create_valid_user_course_invalid_mode`?

eox_core/api/v1/tests/integration/test_views.py
```diff
@@ -370,6 +431,595 @@ def test_update_user_in_tenant_user_not_found(self, param: str, value: str) -> N
...
+        }
+
+        response = self.update_enrollment(self.tenant_x, data=enrollment_data)
+
+        response_data = response.json()
+        self.assertEqual(response.status_code, status.HTTP_202_ACCEPTED)
+        self.assertEqual(response_data["error"]["detail"], f"No enrollment found for {user_data['username']}")
+
+    @ddt.data("email", "username")
+    def test_update_valid_enrollment_force_post(self, param: str) -> None:
```
mariajgrimaldi, line 988: Same question here about what's the difference between updating and force updating.

mariajgrimaldi (changes requested):

eox_core/api/v1/tests/integration/test_views.py
```diff
@@ -370,6 +431,595 @@ def test_update_user_in_tenant_user_not_found(self, param: str, value: str) -> N
...
+        self.assertEqual(response_data["username"], user_data["username"])
+        self.assertEqual(response_data["mode"], enrollment_data["mode"])
+        self.assertEqual(response_data["course_id"], enrollment_data["course_id"])
+        self.assertTrue(response_data["is_active"])
+        self.assertIn("created", response_data)
+
+    @ddt.data("email", "username")
+    def test_get_enrollment_success(self, param: str) -> None:
+        """
+        Test getting a enrollment with a valid user, course and mode in a tenant.
```
mariajgrimaldi, line 720:
```suggestion
        Test getting an enrollment with a valid user, course and mode in a tenant.
```

eox_core/api/v1/tests/integration/test_views.py
```diff
@@ -370,6 +431,595 @@ def test_update_user_in_tenant_user_not_found(self, param: str, value: str) -> N
...
+        self.assertEqual(response_data["username"], user_data["username"])
+        self.assertEqual(response_data["mode"], enrollment_data["mode"])
+        self.assertEqual(response_data["course_id"], enrollment_data["course_id"])
+        self.assertTrue(response_data["is_active"])
+        self.assertIn("created", response_data)
+
+    @ddt.data("email", "username")
+    def test_get_enrollment_does_not_exist(self, param: str) -> None:
+        """
+        Test getting a enrollment that does not exist in a tenant.
```
mariajgrimaldi, line 752:
```suggestion
        Test getting an enrollment that does not exist in a tenant.
```

eox_core/api/v1/tests/integration/test_views.py
```diff
@@ -370,6 +431,595 @@ def test_update_user_in_tenant_user_not_found(self, param: str, value: str) -> N
...
+        response = self.get_enrollment(self.tenant_x, data=enrollment_data)
+
+        response_data = response.json()
+        self.assertEqual(response.status_code, status.HTTP_404_NOT_FOUND)
+        self.assertEqual(response_data, [f"No enrollment found for user:`{user_data['username']}`"])
+
+    @ddt.data("email", "username")
+    def test_get_enrollment_not_found_in_tenant(self, param: str) -> None:
+        """
+        Test getting a enrollment that belongs to another tenant.
```
mariajgrimaldi, line 778:
```suggestion
        Test getting an enrollment that belongs to another tenant.
```

eox_core/api/v1/tests/integration/test_views.py
```diff
@@ -370,6 +431,595 @@ def test_update_user_in_tenant_user_not_found(self, param: str, value: str) -> N
...
+        self.assertIn("detail", response_data)
+        self.assertEqual(
+            response_data["detail"],
+            f"No user found by {{'{param}': '{user_data[param]}'}} on site {self.tenant_y['domain']}.",
+        )
+
+    @ddt.data("email", "username")
+    def test_delete_enrollment_success(self, param: str) -> None:
+        """
+        Test deleting a enrollment with a valid user, course and mode in a tenant.
```
mariajgrimaldi, line 809:
```suggestion
        Test deleting an enrollment with a valid user, course and mode in a tenant.
```

eox_core/api/v1/tests/integration/test_views.py
```diff
@@ -370,6 +431,595 @@ def test_update_user_in_tenant_user_not_found(self, param: str, value: str) -> N
...
+        self.create_enrollment(self.tenant_x, enrollment_data)
+
+        response = self.delete_enrollment(self.tenant_x, data=enrollment_data)
+
+        self.assertEqual(response.status_code, status.HTTP_204_NO_CONTENT)
+
+    @ddt.data("email", "username")
+    def test_delete_enrollment_does_not_exist(self, param: str) -> None:
+        """
+        Test deleting a enrollment that does not exist in a tenant.
```
mariajgrimaldi, line 835:
```suggestion
        Test deleting an enrollment that does not exist in a tenant.
```

eox_core/api/v1/tests/integration/test_views.py
```diff
@@ -370,6 +431,595 @@ def test_update_user_in_tenant_user_not_found(self, param: str, value: str) -> N
...
+        self.assertIn("detail", response_data)
+        self.assertEqual(
+            response_data["detail"],
+            f"No enrollment found for user: `{user_data['username']}` on course_id `{self.course_id}`",
+        )
+
+    @ddt.data("email", "username")
+    def test_delete_invalid_enrollment_for_tenant(self, param: str) -> None:
+        """
+        Test deleting a enrollment that belongs to another tenant.
```
mariajgrimaldi, line 865:
```suggestion
        Test deleting an enrollment that belongs to another tenant.
```

mariajgrimaldi (commented):
I'll leave this so you can commit them in bulk.

eox_core/api/v1/tests/integration/test_views.py
BryanttV: Yes, the test is doing it. When we use ` enrollment_data.pop(param)`

eox_core/api/v1/tests/integration/test_views.py
BryanttV: When we use `force=True` the enrollment is created even if the mode is unavailable in the course.

eox_core/api/v1/tests/integration/test_views.py
BryanttV: The difference is explained here. To update the initial data we must use `force = True`
https://github.com/eduNEXT/eox-core/blob/2708af49afb32ce19bbcd02a17c1d87e24846af8/eox_core/api/v1/views.py#L599-L602

eox_core/api/v1/tests/integration/test_views.py
mariajgrimaldi: It considers only the username but not when both are missing. Am I correct?

eox_core/api/v1/tests/integration/test_views.py
mariajgrimaldi: Can we clarify that in the docstring?

.github/workflows/integration-test.yml
```diff
@@ -7,7 +7,11 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        tutor_version: ['<18.0.0', '<19.0.0']
+        include:
```
magajh, line 10: @BryanttV can we please explain why is this implementation necessary in the PR description? I think it's important to explain the reasoning behind this decision

-------------------------
Since I’m considering moving `tutor local do importdemocourse` to the GitHub Action, should we also pass the course ID to the environment in the step that runs the integration tests? This way, the action would handle the full demo course setup, and we can ensure that the tests always reference the correct course and prevent each plugin from setting it manually

I was able to obtain the course ID dynamically with:
```bash
COURSE_KEY=$(tutor local exec lms python manage.py lms shell -c "from openedx.core.djangoapps.content.course_overviews.models import CourseOverview; course = CourseOverview.objects.first(); print(course.id, end='')")
```

However, if the import process changes in future Open edX versions, this could break the tests. Of course, we could always introduce version-based logic to handle that, but we could also just set the course ID manually in the action (assuming the Demo course ID won’t change frequently). Starting from Tutor version 18.0.0, the ID is `course-v1:OpenedX+DemoX+DemoCourse`, so we’d handle logic for both IDs for now, and remove support for versions below 18.0.0 later

I’d like to hear your thoughts on this approach @BryanttV @mariajgrimaldi

eox_core/api/v1/tests/integration/test_views.py
BryanttV: At least one (username or email) must be sent for the request to work. In this case, I did not include the email since we want to test when neither of these two fields is sent.

eox_core/api/v1/tests/integration/test_views.py
BryanttV: This test is not only intended to verify that we can send the `force = True` flag without any problem. We could unify the tests and include this flag inside the `ddt.data`: https://github.com/eduNEXT/eox-core/commit/5c42efe848e36dc760d20bfe0e6066d945edaa13

eox_core/api/v1/tests/integration/test_views.py
BryanttV: I included the `param` and `value` in the test parameters.

Makefile
```diff
@@ -50,7 +50,7 @@ python-quality-test:
 run-tests: python-test python-quality-test
 
 run-integration-tests: install-dev-dependencies
-	pytest ./eox_core --ignore-glob='**/unit/*' --ignore-glob='**/edxapp_wrapper/*'
+	pytest -rP ./eox_core --ignore-glob='**/unit/*' --ignore-glob='**/edxapp_wrapper/*'
```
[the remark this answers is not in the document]
BryanttV: Thanks for your suggestion! https://github.com/eduNEXT/eox-core/commit/9aa144a3fc620da14081e102e08ba6a026dc0415

eox_core/api/v1/tests/integration/test_views.py
BryanttV: Done: https://github.com/eduNEXT/eox-core/commit/490a19eee770c832882cb730a705843911370df9

eox_core/api/v1/tests/integration/test_views.py
BryanttV: Done: https://github.com/eduNEXT/eox-core/commit/490a19eee770c832882cb730a705843911370df9

eox_core/api/v1/tests/integration/test_views.py
BryanttV: It occurs to me that we can query that the enrollment was created by doing a `self.get_enrollment()` and verify that code **200** or **404** is returned, what do you say?

.github/workflows/integration-test.yml
BryanttV: I really like what you propose. I think it's more appropriate to leave that responsibility to the action, than to keep it in every plugin.

eox_core/api/v1/tests/integration/test_views.py
BryanttV: The problem is that we need the user's data in each test. I'm not sure how we would access it afterwards. Do you have any idea?

BryanttV: I think I know how to do it, I'll do some tests.

BryanttV: Done: https://github.com/eduNEXT/eox-core/commit/bd1d3c1ef1c7c0aea3864dbe491df456a0b0b43f

eox_core/api/v1/tests/integration/test_views.py
mariajgrimaldi: This is the test case I'm referring to: 

```
curl -X 'POST' \
  'https://redwood.../eox-core/api/v1/enrollment/' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -H 'X-CSRFToken: ...' \
  -d '{
  "force": false,
  "is_active": true,
  "mode": "audit",
  "enrollment_attributes": [],
  "course_id": "course-v1:OpenedX+DemoX+DemoCourse",
  "created": "2021-10-11T10:06:17.876505-05:00"
}'
```
Can we include it as well?

eox_core/api/v1/tests/integration/test_views.py
mariajgrimaldi: Yes, that would align with what we agreed on in the document.

.github/workflows/integration-test.yml
mariajgrimaldi: We should definitely move this to the GH action itself.

I like the proposal's automatic approach to getting the course ID. However, I'm very wary of using code from the platform without any API or interface, even from the shell. I don't think it's worth it, considering the course ID hasn't changed much over time.

mariajgrimaldi (dismissed):
I'll leave my approval again once we address the last couple of comments. Thank you!

eox_core/api/v1/tests/integration/test_views.py
BryanttV: I updated the test of enrollment's API: https://github.com/eduNEXT/eox-core/commit/69692fa07ef30e5dffdc668ef586b731ab72f569
> Do you think we should consider updating the user's API as well as the next step? I think we missed doing those checks there.

Yes, we should also update the user's API tests. I will create a task for it.

scripts/execute_integration_tests.sh
```diff
@@ -1,2 +1,3 @@
 #!/bin/bash
+tutor local do importdemocourse
```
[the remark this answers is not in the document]
BryanttV: Done: https://github.com/eduNEXT/eox-core/commit/658b2fbf443af1ca1c148b4fe53322d41a83282b

.github/workflows/integration-test.yml
```diff
@@ -7,7 +7,11 @@ jobs:
...
+          - tutor_version: '<18.0.0'
+            course_id: 'course-v1:edX+DemoX+Demo_Course'
+          - tutor_version: '<19.0.0'
+            course_id: 'course-v1:OpenedX+DemoX+DemoCourse'
```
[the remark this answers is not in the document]
BryanttV: Done: https://github.com/eduNEXT/eox-core/commit/a7bae952580f4da2a3a4badf241cc624084138e3

eox_core/api/v1/tests/integration/test_views.py
```diff
@@ -16,6 +18,75 @@
 
 settings = ds.INTEGRATION_TEST_SETTINGS
 
+USER_URL = f"{settings['EOX_CORE_API_BASE']}{reverse('eox-api:eox-api:edxapp-user')}"
+USER_UPDATER_URL = f"{settings['EOX_CORE_API_BASE']}{reverse('eox-api:eox-api:edxapp-user-updater')}"
+ENROLLMENT_URL = f"{settings['EOX_CORE_API_BASE']}{reverse('eox-api:eox-api:edxapp-enrollment')}"
```
mariajgrimaldi, line 23: This is not a blocker in any way. However, API URLs are usually defined inside each test suite setup method. Is there a specific reason for defining them here?

BryanttV: I tried to define the URLs in each `SetUp` method but generated an error with the variables due to multiple inheritance. However, we can add these URLs in each Mixin: https://github.com/eduNEXT/eox-core/commit/a42c18622d9b24d566c7286f58a0a5ec887ff664

What do you think?

mariajgrimaldi: That looks reasonable. Thank you!

magajh (approved):
Thanks for addressing the comments!

PR #292, opened 2024-10-22
chore: drop support for Python 3.10 and Django 3.2

magajh:
**Description**:

This PR removes support for Django 3.2 and Python 3.10 from our plugin.

**Rationale**:

We have decided to drop support for these versions because none of the Open edX releases we support use Django 3.2 or Python 3.10. Therefore, maintaining compatibility with them is unnecessary.

For more context and information, please refer to [AP-1490](https://edunext.atlassian.net/browse/AP-1490).

**Changes**:

- Updated the test workflow to remove Python 3.10 and Django 3.2 from the Python and Django matrix.
- Updated the `tox.ini` file to eliminate specifications for Python 3.10 and Django 3.2. 
- Deleted the `django32.txt` file.
- Removed constraints that were only necessary for compatibility with Python 3.10 or Django 3.2.
- Upgraded the requirements using Python 3.8, which is the oldest version we now support.

**How to Test**:

- Ensure that all tests in the GitHub CI pipeline pass successfully.
- Verify that the pipeline no longer runs tests with Python 3.10 and Django 3.2.

[AP-1490]: https://edunext.atlassian.net/browse/AP-1490?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BryanttV (commented):
Thanks @magajh! Should we update changelog and bump version?

MaferMazu:
I think we will need the bump version, and a major bump version. Because in our readme, we said that the latest version of eox-core works on maple, nutmeg, olive, and palm in the [compatibilities notes](https://github.com/eduNEXT/eox-core?tab=readme-ov-file#compatibility-notes). And with this change, for the dropping support of Django 3.2, that is [the Django version used palm](https://github.com/openedx/edx-platform/blob/open-release/palm.master/requirements/edx/base.txt), we are dropping support for the releases before or equals palm.

And based on [Release Strategy doc](https://internal.docs.edunext.co/en/latest/internal/Teams/Descontrol/Internal_processes/releases_strategy.html), we should use a major if we change the compatibility and require users to adapt (in this case, someone using palm could no longer use the latest release of eox-core).

I understand this is not a bump version change if we only think in Python and Django, but thinking about the release we support of Open edX, we are dropping support for maple, nutmeg, olive, and palm, which should be a major change.

So, we probably want to update the readme to have an upper limit for those releases we are dropping support for and apply a major version.

What do you think?

BryanttV:
I was going to comment with that making a patch was the best thing to do but... not anymore, I agree with @MaferMazu's comment.

CHANGELOG.md
```diff
@@ -5,13 +5,19 @@ All notable changes to this project will be documented in this file.
...
 <!-- Content should be placed here -->
 
-## [v10.6.0](https://github.com/eduNEXT/eox-core/compare/v10.5.1...v10.6.0) - (2024-09-17)
+## [v11.0.0](https://github.com/eduNEXT/eox-core/compare/v10.6.0...v11.0.0) - (2024-10-22)
+
+#### ⚠ BREAKING CHANGES
+
+- **Dropped Support for Django 3.2**: Removed support for Django 3.2 in this plugin. As a result, we have also dropped support for Open edX releases from Maple up to and including Palm, which rely on Django 3.2. Future versions of this plugin may not be compatible with these Open edX releases.
+
+## [v10.6.0](https://gitshub.com/eduNEXT/eox-core/compare/v10.5.1...v10.6.0) - (2024-09-17)
```
BryanttV, line 20:
```suggestion
## [v10.6.0](https://github.com/eduNEXT/eox-core/compare/v10.5.1...v10.6.0) - (2024-09-17)
```

magajh: thanks!

BryanttV (approved):
LGTM!

MaferMazu (approved):
It looks good to me. Thanks a lot for this ✨

PR #298, opened 2024-10-28
test: add TestPathRedirectionMiddleware integration tests `AP-1547`

BryanttV:
## Description
This PR adds integration tests for the `TestPathRedirectionMiddleware` middleware. These tests are executed in the job of **Tutor Integration Tests**

## Testing instructions

## Jira Issue
- [AP-1547](https://edunext.atlassian.net/browse/AP-1547)

[AP-1547]: https://edunext.atlassian.net/browse/AP-1547?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


eox_core/tests/integration/test_middleware.py
```diff
@@ -0,0 +1,58 @@
...
+
+settings = ds.INTEGRATION_TEST_SETTINGS
+
+ACCESS_TOKEN = get_access_token()
+
+
+class TestPathRedirectionMiddleware(BaseIntegrationTest):
+    """Integration tests for the PathRedirectionMiddleware."""
+
+    def setUp(self):
```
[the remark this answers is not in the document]
BryanttV: Yes, I added new tests: https://github.com/eduNEXT/eox-core/commit/ab56e371b14e6ac23d242826006a07ed7782f613

eox_core/tests/integration/utils.py
```diff
@@ -0,0 +1,108 @@
...
+        timeout=settings["API_TIMEOUT"],
+    )
+
+
+class BaseIntegrationTest(TestCase):
+    """
+    Base class for the integration test suite.
+    """
+
+    def __init__(self, *args, **kwargs):
```
mariajgrimaldi, line 82: Out of curiosity, again. Why are we using `__init__` instead of the standard `setup` for test case classes?

eox_core/tests/integration/utils.py
```diff
@@ -0,0 +1,108 @@
...
+
+        Args:
+            prefix (str, optional): The tenant prefix. Defaults to "".
+
+        Returns:
+            dict: The tenant data.
+        """
+        domain = f"{prefix}.{settings['LMS_BASE']}" if prefix else settings["LMS_BASE"]
+        return {
+            "base_url": f"http://{domain}",
```
mariajgrimaldi, line 106: Is this always going to be http, right?

mariajgrimaldi (approved):
I left a few non-blocking comments for you to address. Other than that, it looks good to me!

eox_core/tests/integration/utils.py
```diff
@@ -0,0 +1,108 @@
...
+        "client_id": settings["CLIENT_ID"],
+        "client_secret": settings["CLIENT_SECRET"],
+        "grant_type": "client_credentials",
+    }
+    url = f"http://{settings['LMS_BASE']}/oauth2/access_token/"
+    response = requests.post(url, data=data, timeout=settings["API_TIMEOUT"])
+    return response.json()["access_token"]
+
+
+ACCESS_TOKEN = get_access_token()
```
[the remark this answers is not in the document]
BryanttV: This is to avoid generating a new token for each request. Otherwise, an error is raised due to many requests being made.

eox_core/tests/integration/utils.py
BryanttV: This is mainly done to make the `self.tenant_x_url` attribute available in all methods, such as the `get_session` method. If we use `setUp,` the attribute would only be available for test methods.

eox_core/tests/integration/utils.py
BryanttV: Yes, always. The Tutor environment in the integration tests action runs with http

mariajgrimaldi:
@BryanttV, thanks for the context!

mariajgrimaldi (approved):

plugins/eox-core.yml
magajh, line 1: I'm unsure about these paths and the plugin file name. I think we should consider renaming this folder and file to better reflect their purpose—perhaps something like integration_tests/tutor_plugins/openedx-common-settings.yml, though I'm not entirely sure about this either.

Do you have any other ideas on how we could better organize or store these Tutor plugins within the codebase?

magajh (commented):
Thank you @BryanttV! I just have a couple of comments

plugins/eox-core.yml
BryanttV: Yes, it could have a more appropriate name, or maybe `middleware-settings.yml`. That gets me thinking about having a folder called `integration_tests` and leaving the `scripts` and `tutor_plugins` folders there. What do you think?
```bash
integration_tests/
├── scripts/
│   ├── execute_integration_tests.hs
│   └── execute_tutor_extra_commands.sh
└── tutor_plugins/
    └── middleware-settings.yml
```

magajh: @BryanttV I like this approach. Keeping everything related to the integration tests workflow in a single folder makes it more organized and easier to locate

BryanttV: Thanks, @magajh, I made the changes and executed the integration test workflow with the new paths: https://github.com/eduNEXT/eox-core/actions/runs/11898739065/job/33155867798?pr=298

magajh (approved):

PR #300, opened 2024-10-30
test: add integration tests for the support api `AP-1527`

BryanttV:
## Description
This PR adds integration tests for the support API. These tests are executed in the job of **Tutor Integration Tests**

## Testing instructions

## Jira Issue
- [AP-1527](https://edunext.atlassian.net/browse/AP-1527)

[AP-1527]: https://edunext.atlassian.net/browse/AP-1527?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.coveragerc
magajh, line 1: Isn't this configuration already in https://github.com/eduNEXT/eox-core/blob/master/setup.cfg#L11-L21?

eox_core/api/support/v1/tests/integration/test_views.py
```diff
@@ -0,0 +1,294 @@
...
+            data (dict, optional): The body data for the request.
+
+        Returns:
+            requests.Response: The response object.
+        """
+        return make_request(tenant, "POST", url=self.OAUTH_APP_URL, json=data)
+
+
+@ddt.ddt
+class TestEdxAppUserAPIIntegration(SupportAPIRequestMixin, BaseIntegrationTest, UsersAPIRequestMixin):
```
magajh, line 71: I know this PR is still a draft, so I’m holding off on a full review for now, but I think the formatting would look cleaner with:

```suggestion
class TestEdxAppUserAPIIntegration(
    SupportAPIRequestMixin,
    BaseIntegrationTest,
    UsersAPIRequestMixin,
):
```

.coveragerc
BryanttV: Yes, you're right! I deleted the `.coveragerc` file: https://github.com/eduNEXT/eox-core/commit/10a050b96e4b3fb8095e05aba41292798e790170, ty

eox_core/api/support/v1/tests/integration/test_views.py
BryanttV: Done: https://github.com/eduNEXT/eox-core/commit/3ff25e8699849a907f7c6e184d18b8d7ee45e032

eox_core/api/support/v1/tests/integration/test_views.py
```diff
@@ -0,0 +1,298 @@
...
+
+        Open edX definitions tested:
+        - `get_edxapp_user`
+
+        Expected result:
+        - The status code is 404.
+        - The response indicates the user was not found in the tenant.
+        """
+        response = self.delete_user(self.tenant_x, {query_param: value})
+        response_data = response.json()
```
mariajgrimaldi, line 125: What if I try to delete a user from a different tenant? Should that work?

eox_core/api/support/v1/tests/integration/test_views.py
```diff
@@ -0,0 +1,298 @@
...
+        Expected result:
+        - The status code is 200.
+        - The response indicates the username was updated successfully.
+        - The user is found in the tenant with the new username.
+        """
+        data = next(FAKE_USER_DATA)
+        self.create_user(self.tenant_x, data)
+        new_username = f"new-username-{query_param}"
+
+        response = self.update_username(self.tenant_x, {query_param: data[query_param]}, {"new_username": new_username})
```
mariajgrimaldi, line 167: Same question here about updating a user from a different tenant. Should that work?

eox_core/api/support/v1/tests/integration/test_views.py
```diff
@@ -0,0 +1,298 @@
...
+        - `create_edxapp_user`
+        - `UserSignupSource`
+
+        Expected result:
+        - The status code is 200.
+        - The response indicates the oauth application was created successfully.
+        """
+        user_data = next(FAKE_USER_DATA)
+        if create_user:
+            self.create_user(self.tenant_x, user_data)
```
mariajgrimaldi, line 253: Could you explain in the docstring why creating the user is needed?

mariajgrimaldi (changes requested):

eox_core/api/support/v1/tests/integration/test_views.py
BryanttV: The endpoint creates the user if the user doesn't exist. The conditional is to validate both cases. I updated the docstring: https://github.com/eduNEXT/eox-core/commit/bb37aab1015654d52b843dfae5ecc542dc8f9576

eox_core/api/support/v1/tests/integration/test_views.py
BryanttV: I added a test for this: https://github.com/eduNEXT/eox-core/commit/47e03c993866e0d63471354a2d0b1f2469e1aa71

eox_core/edxapp_wrapper/backends/users_m_v1.py
```diff
@@ -281,7 +281,7 @@ def delete_edxapp_user(*args, **kwargs):
     signup_sources = user.usersignupsource_set.all()
     sources = [signup_source.site for signup_source in signup_sources]
 
-    if site and site.name.upper() in (source.upper() for source in sources):
+    if site and site.domain.upper() in (source.upper() for source in sources):
```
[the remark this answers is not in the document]
mariajgrimaldi: But I don't think we should change this directly in Maple's backend, mainly because we're testing with quince, redwood, and nightly. Should we rename this backend to be compatible with the Redwood release instead?

mariajgrimaldi:
LGTM, I have an additional comment for you to address: https://github.com/eduNEXT/eox-core/pull/300#discussion_r1836465656. Let me know once you do to leave my approval :)

eox_core/edxapp_wrapper/backends/users_m_v1.py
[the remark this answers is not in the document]
BryanttV: Yes, we should rename the backend but, I tested in Quince, and the same thing happens. I'm not sure which version has the same behavior. How should we proceed in this case?

scripts/execute_tutor_extra_commands.sh
```diff
@@ -0,0 +1,2 @@
+# Populate the retirement states
+tutor local exec lms python manage.py lms populate_retirement_states
```
magajh, line 2: Would it be possible to just run this within the script that executes the test command and skip this script entirely?
[https://github.com/eduNEXT/integration-test-in-tutor/pull/8/files#r183854869](https://github.com/eduNEXT/integration-test-in-tutor/pull/8/files#r1838548698)

.github/workflows/integration-test.yml
```diff
@@ -7,14 +7,16 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        tutor_version: ['<18.0.0', '<19.0.0', 'nightly']
+        tutor_version: ['<18.0.0', '<19.0.0']
```
magajh, line 10: remember to undo this change before merging please

magajh (commented):
Thank you for this @BryanttV! just have one comment about the extra commands script

magajh (dismissed):
@BryanttV tag me if you need my approval again

eox_core/api/support/v1/tests/integration/test_views.py
```diff
@@ -0,0 +1,354 @@
...
+        Test delete an edxapp user in a tenant without providing the username or email.
+
+        Expected result:
+        - The status code is 400.
+        - The response indicates the username or email is required.
+        """
+        response = self.delete_user(self.tenant_x)
+        response_data = response.json()
+
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
```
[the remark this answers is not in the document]
BryanttV: I don't think it's necessary to check it here, anyway in those tests no users are being created.

eox_core/edxapp_wrapper/backends/users_m_v1.py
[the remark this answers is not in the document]
mariajgrimaldi: Mmm. Then, since we still support quince, let's use q instead since it's the oldest version we support and we know that this backend works. What do you think?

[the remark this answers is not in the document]
BryanttV: Yes, I'm agree. https://github.com/eduNEXT/eox-core/commit/cca23f4e2205ae8018044994859812f7f1cb8d5a

BryanttV:
Hi @magajh @mariajgrimaldi, I made the last changes in the PR.

eox_core/api/support/v1/tests/integration/test_views.py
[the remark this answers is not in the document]
magajh: I understand that we aren’t creating any users in these tests, but is it possible that users created in previous tests might still exist? If I’m not mistaken, these tests don’t reset the database after each test case, so any user created would persist unless explicitly deleted. That’s why I think it could make sense to ensure our delete operation doesn’t unintentionally remove any users.

That said, this type of check might not be necessary, as it’s not directly testing an integration scenario. Still, it’s something we can think about further to define the boundaries of database state verification. For now, though, I think it’s fine as is

magajh (approved):

eox_core/api/support/v1/tests/integration/test_views.py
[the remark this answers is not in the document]
mariajgrimaldi: @magajh: Now that you mention it, should we consider using a teardown method to undo all operations made during the tests in the near future? This way, we know running a test suite is an idempotent operation.

mariajgrimaldi (approved):

PR #303, opened 2024-11-08
chore: removed unnecessary constraints

Asespinel:
**Description**

This PR removes the pycodestyle as the issue was already fixed [here](https://github.com/PyCQA/pycodestyle/issues/1090) and added a comment explaing why the django-oauth-toolkit<2.0.0 is necessary.

### DIFF
--- a/eox_core/api/v1/serializers.py
+++ b/eox_core/api/v1/serializers.py
@@ -216,6 +216,10 @@ class WrittableEdxappUserSerializer(EdxappExtendedUserSerializer):
         has_profile = hasattr(instance, 'profile')
 
         extra_registration_fields.update(extended_profile_fields)
+
+        # Initialize profile_meta to avoid potential unassigned variable warning
+        profile_meta = {}
+        
         if has_profile:
             profile_meta = instance.profile.get_meta()
 
--- a/eox_core/api/v1/serializers.py
+++ b/eox_core/api/v1/serializers.py
@@ -44,6 +44,7 @@ class EdxappWithWarningSerializer(serializers.Serializer):
     """
     Mixin serializer to add a warning field to Edxapp serializers
     """
+
     def __init__(self, *args, **kwargs):
         """
         Conditionally adds a warning field if a context is passed
@@ -219,7 +220,7 @@ class WrittableEdxappUserSerializer(EdxappExtendedUserSerializer):
 
         # Initialize profile_meta to avoid potential unassigned variable warning
         profile_meta = {}
-        
+
         if has_profile:
             profile_meta = instance.profile.get_meta()
 
@@ -295,6 +296,7 @@ class EdxappValidatedCourseIDField(serializers.Field):
     """
     CourseKey Field
     """
+
     def to_representation(self, value):
         return str(value)
 
--- a/requirements/constraints.txt
+++ b/requirements/constraints.txt
@@ -15,7 +15,6 @@
 # Ensures a consistent pylint version across all Python environments to avoid triggering varying warnings
 pylint<3.3.0
 
-# Date: 2023-07-31
 # As it is not clarified what exact breaking changes will be introduced as per
 # the next major release, ensure the installed version is within boundaries.
 # Issue for unpinning: https://github.com/openedx/edx-platform/issues/32884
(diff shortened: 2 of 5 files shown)


requirements/constraints.txt
```diff
@@ -11,9 +11,11 @@
...
 -c https://raw.githubusercontent.com/openedx/edx-lint/master/edx_lint/files/common_constraints.txt
 
-# Keep same platform version
-pylint<3.0
-pycodestyle<2.9.0
+# Versions >= 3.3.0 drop support for Python 3.8
+# Ensures a consistent pylint version across all Python environments to avoid triggering varying warnings
+pylint<3.3.0
+
+# Higher version has breaking changes in the platform
```
BryanttV, line 18: Could we give a little more detail as it is here? https://github.com/openedx/edx-platform/blob/31d170a687795a7bda1e7b510d03d08fd00e1f22/requirements/constraints.txt#L48-L52

Asespinel: Done !

requirements/constraints.txt
```diff
@@ -11,9 +11,11 @@
...
 -c https://raw.githubusercontent.com/openedx/edx-lint/master/edx_lint/files/common_constraints.txt
 
-# Keep same platform version
-pylint<3.0
-pycodestyle<2.9.0
+# Versions >= 3.3.0 drop support for Python 3.8
+# Ensures a consistent pylint version across all Python environments to avoid triggering varying warnings
+pylint<3.3.0
+
+# Higher version has breaking changes in the platform as described here https://github.com/openedx/edx-platform/issues/35277
```
mariajgrimaldi, line 18: I think we should start adding when this constraint should be dropped as part of this comment if the issue linked doesn't offer any other context. In this case, issue/35277 clearly says when the constraint is being unpinned, but when should we unpin pylint? I don't want this to be a blocker, but we need to offer more details when adding restrictions so we don't have to guess whether a constraint should be dropped.

eox_core/api/v1/serializers.py
```diff
@@ -216,6 +217,10 @@ def update(self, instance, validated_data):
         has_profile = hasattr(instance, 'profile')
 
         extra_registration_fields.update(extended_profile_fields)
+
+        # Initialize profile_meta to avoid potential unassigned variable warning
+        profile_meta = {}
+
```
mariajgrimaldi, line 223: Can we remove line 235? Also, I'm guessing this is covered by unittests, right?

mariajgrimaldi (changes requested):

requirements/constraints.txt
mariajgrimaldi: Actually, this is the format that's used in edx-platform:

```
# Date: 2020-02-26
# As it is not clarified what exact breaking changes will be introduced as per
# the next major release, ensure the installed version is within boundaries.
# Issue for unpinning: https://github.com/openedx/edx-platform/issues/35280
```
Should we consider adopting it?

eox_core/api/v1/serializers.py
Asespinel: Hi majo, I had to add that line since a test was failing with the message: E0606: Possibly using variable 'profile_meta' before assignment (possibly-used-before-assignment).

requirements/constraints.txt
Asespinel: Sure, I'll add more info on the constraint and it probablya  good idea to have the same syntax across all repos.

requirements/constraints.txt
```diff
@@ -11,9 +11,14 @@
...
 -c https://raw.githubusercontent.com/openedx/edx-lint/master/edx_lint/files/common_constraints.txt
 
-# Keep same platform version
-pylint<3.0
-pycodestyle<2.9.0
+# Versions >= 3.3.0 drop support for Python 3.8
+# Ensures a consistent pylint version across all Python environments to avoid triggering varying warnings
+pylint<3.3.0
+
+# Date: 2023-07-31
```
mariajgrimaldi, line 18: Thank you for considering my suggestion! The most relevant parts of this format are the description and the issue, if it exists and any reference to when to unpin the constraint. So I think we can drop the date. We could still do a git blame to check when it was introduced.

mariajgrimaldi (approved):

mariajgrimaldi (approved):

BryanttV (approved):

PR #308, opened 2024-11-26
ci: set fail-fast as False to allow integration tests to continue on nightly Tutor failures

magajh:
## Description

This PR updates the Tutor Integration Tests workflow to ensure that failures in the nightly version of Tutor do not cancel the execution of tests for stable versions (<18.0.0 and <19.0.0).  `fail-fast: false` is added to the strategy, allowing the workflow to proceed even if nightly fails.

This change is important because the nightly version of Tutor is not stable and may occasionally fail due to issues unrelated to our tests. By implementing this update, we ensure that failures in nightly do not block critical tests for stable versions.

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -21,3 +21,5 @@ jobs:
           fixtures_file: 'fixtures/initial_data.json'
           openedx_imports_test_file_path: 'eox_core/edxapp_wrapper/tests/integration/test_backends.py'
           tutor_extra_commands_path: 'integration_tests/scripts/execute_tutor_extra_commands.sh'
+        continue-on-error: ${{ matrix.tutor_version == 'nightly' }}
+
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -21,5 +21,3 @@ jobs:
           fixtures_file: 'fixtures/initial_data.json'
           openedx_imports_test_file_path: 'eox_core/edxapp_wrapper/tests/integration/test_backends.py'
           tutor_extra_commands_path: 'integration_tests/scripts/execute_tutor_extra_commands.sh'
-        continue-on-error: ${{ matrix.tutor_version == 'nightly' }}
-
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -7,7 +7,10 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        tutor_version: ['<19.0.0', '<20.0.0', 'nightly']
+        tutor_version: ['<18.0.0', '<19.0.0']
+        include:
+          - tutor_version: 'nightly'
+            continue_on_error: true
     steps:
       - name: Run Integration Tests
         uses: eduNEXT/integration-test-in-tutor@main
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -6,11 +6,9 @@ jobs:
     name: Tutor Integration Tests
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
-        tutor_version: ['<18.0.0', '<19.0.0']
-        include:
-          - tutor_version: 'nightly'
-            continue_on_error: true
+        tutor_version: ['<19.0.0', '<20.0.0', 'nightly']
     steps:
       - name: Run Integration Tests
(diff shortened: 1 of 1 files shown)


magajh:
For more context on this solution, see https://github.com/eduNEXT/eox-tenant/pull/232#issuecomment-2500879080

BryanttV (commented):
Thanks @magajh! This change affects all Tutor versions, right? For example, if v18 integration tests fail, do the other versions still run?

magajh:
> Thanks @magajh! This change affects all Tutor versions, right? For example, if v18 integration tests fail, do the other versions still run?

That's correct @BryanttV. I couldn't find a way to continue the execution for the other tutor_versions in the matrix only if 'nightly' fails. The use of `continue-on-error` marks the workflow for nightly as passed even if something fails during the execution

BryanttV:
@magajh, Could you try this? I don't know if it works
```yaml
jobs:
  integration-test:
    name: Tutor Integration Tests
    runs-on: ubuntu-latest
    continue-on-error: ${{ matrix.experimental }}
    strategy:
      fail-fast: false
      matrix:
        tutor_version: ['<18.0.0', '<19.0.0']
        experimental: [false]
        include:
          - version: 'nightly'
            experimental: true
```
I was guided by [this docs](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast). However, if it doesn't work or we don't want to complicate things, it seems to me that the current solution is a good one.

magajh:
> @magajh, Could you try this? I don't know if it works
> 
> ```yaml
> jobs:
>   integration-test:
>     name: Tutor Integration Tests
>     runs-on: ubuntu-latest
>     continue-on-error: ${{ matrix.experimental }}
>     strategy:
>       fail-fast: false
>       matrix:
>         tutor_version: ['<18.0.0', '<19.0.0']
>         experimental: [false]
>         include:
>           - version: 'nightly'
>             experimental: true
> ```
> 
> I was guided by [this docs](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast). However, if it doesn't work or we don't want to complicate things, it seems to me that the current solution is a good one.

@BryanttV Thanks for the suggestion! Since we’re only applying `continue-on-error` for the nightly version, I think we don’t need to introduce the additional `experimental` field in the matrix. That field could be useful for handling things dynamically across multiple versions, but in this case, looks like it adds unnecessary complexity.

Using `continue-on-error: ${{ matrix.tutor_version == 'nightly' }}` should be a simpler and more direct way to achieve the same result.

That said, one limitation of `continue-on-error` is that it marks the nightly tests as passed even if they fail. If that’s not what we want, we might need to explore a different approach to ensure the failures are visible while still allowing the workflow to continue, as it happened with the last commit https://github.com/eduNEXT/eox-core/pull/308/commits/dc479c75f51a1990144262373a21b2f953def7fc

.github/workflows/integration-test.yml
```diff
@@ -21,3 +21,5 @@ jobs:
           fixtures_file: 'fixtures/initial_data.json'
           openedx_imports_test_file_path: 'eox_core/edxapp_wrapper/tests/integration/test_backends.py'
           tutor_extra_commands_path: 'integration_tests/scripts/execute_tutor_extra_commands.sh'
+        continue-on-error: ${{ matrix.tutor_version == 'nightly' }}
```
mariajgrimaldi, line 24: Is the `fail-fast` no longer needed? I see it in the cover letter, so I'm guessing it was dropped.

magajh: @mariajgrimaldi yes `fail-fast ` is what actually works. This `continue-on-error` marks the tests in nightly as successful https://github.com/eduNEXT/eox-core/actions/runs/12036115374/job/33556601073?pr=308
I'll revert the change

magajh:
@mariajgrimaldi @BryanttV @jignaciopm this is ready for the final review

.github/workflows/integration-test.yml
mariajgrimaldi: Good, thank you!

mariajgrimaldi (approved):

BryanttV (approved):
LGTM! Thanks @magajh

BryanttV (approved):

mariajgrimaldi (approved):

PR #313, opened 2025-01-09
feat: change logger level in order to avoid unnecessary sentry notifi…

andrey-canon:
## Description

Change logger level to avoid unnecessary sentry notifications


johanseto (approved):
LGTM

PR #333, opened 2025-12-17
feat: use shared behavior class for updating email and username

mariajgrimaldi:
## Description

Refactor the current username update flow to use a shared behavior class and base updater view and serializer. This enables supporting email updates separately via a different endpoint. 

## Testing instructions

1. Create a user by calling `{{domain}}/eox-core/api/v1/user/` with the following payload:
```
{
   "username": "{{username}}",
   "email": "{{email}}",
   "password": "p@ssword",
   "fullname": "To Be Deleted"
}
```
2. Change their email by calling {{domain}}/eox-core/support-api/v1/user/replace-email/?username={{username}}
```
{
    "new_email": "changed_{{email}}"
}
```
I'll implement integration tests once the pipeline is fixed cause I tried running it locally but failed doing so. 

## Additional information

A client requested the ability to update a user's email address immediately after it is changed in their single sign-on (SSO) identity provider. This is required to maintain GDPR compliance, as communications must always be delivered to the correct and current email recipient. Currently, although the email can be updated via SSO, the change is not applied in the platform until the user logs in. This creates a risk where emails are sent to an outdated address if the user does not log in immediately, potentially resulting in non-compliance with GDPR data management standards.

## Checklist for Merge

- [ ] Updated documentation
- [ ] Rebased master/main
- [ ] Squashed commits


mariajgrimaldi:
@eduNEXT/heimdall this PR is still a work in progress, but we have not been able to get a single successful Tutor integration test run. Do you know if there is something wrong with the pipeline? Thanks in advance!

magajh:
@mariajgrimaldi, the integration test fix has been applied. Could you please rebase this PR with master?

mariajgrimaldi:
@magajh, thanks! I'll be implementing the integration tests for the replace-email endpoint

jignaciopm (dismissed):
LGTM!

jignaciopm (approved):
LGTM!

PR #335, opened 2026-01-02
feat: add flag to allow permanent user deletions

mariajgrimaldi:
## Description

Use a feature flag to allow permanent deletion of a user and their related objects. This path would skip the retirement pipeline, where some PII is anonymized but still prevents a user from registering again with the same email or username. The feature flag would be off by default, so the current deletion behavior is maintained unless an administrator explicitly enables permanent deletion.

I also considered adding a separate delete endpoint to avoid surprising API users once the flag is enabled. A user could call delete user expecting the usual behavior and instead fully remove the account without realizing that permanent deletion is active. The downside is that we would end up with two delete endpoints that look very similar, which could be confusing.

What do you think about this approach?

## Testing instructions

1. Inlcude this configuration in a tutor plugin to enable permanent deletions:

```
from tutor import hooks

hooks.Filters.ENV_PATCHES.add_item(
    (
        "openedx-lms-common-settings",
"""
EOX_CORE_ALLOW_PERMANENT_USER_DELETION = True
"""
    )
)
```
3. Create a user using the API or another mechanism, and then call the delete endpoint as usual:
Screenshot from 2026-01-02 12-42-20

## Additional information
A client was about to use this API. While reviewing it, we noticed a problem. As with the LMS deletion flow, once a user account is deleted, the same username or email cannot be used again.

This goes against the client’s internal policies. It can also raise GDPR concerns, because the system still appears to keep the user’s PII, even if it is anonymized. This is the main reason for allowing administrators to fully delete users.

## Checklist for Merge

- [ ] Updated documentation
- [ ] Rebased master/main
- [ ] Squashed commits

### DIFF
--- a/eox_core/edxapp_wrapper/backends/users_q_v1.py
+++ b/eox_core/edxapp_wrapper/backends/users_q_v1.py
@@ -265,9 +265,41 @@ def get_edxapp_user(**kwargs):
     return user
 
 
-def delete_edxapp_user(*args, **kwargs):
+def permanently_delete_user(*args, **kwargs):
+    """Hard deletes a user from the platform.
+
+    This function permanently removes the user from the database. It is
+    irreversible and should be used with caution. All associated objects should
+    be handled via cascading deletes or other mechanisms as needed. The
+    only object manually deleted here are the OAuth tokens associated with the user.
     """
-    Deletes a user from the platform.
+    user = kwargs.get("user")
+
+    with transaction.atomic():
+        # Delete OAuth tokens associated with the user before deleting the user.
+        # These are not set to cascade delete automatically.
+        retire_dot_oauth2_models(user)
+        UserSocialAuth.objects.filter(user_id=user.id).delete()
+
+        user.delete()
+
+    return (
+        f"The user {user.username} <{user.email}> has been permanently deleted",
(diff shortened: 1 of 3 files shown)


mariajgrimaldi:
I tried to add test cases for the feature flag, but I am not sure how to change Tutor settings during a test run. Because of it I reverted the commit. Is this even feasible with the current setup?

eox_core/edxapp_wrapper/backends/users_q_v1.py
```diff
@@ -265,9 +265,43 @@ def get_edxapp_user(**kwargs):
...
+    user = kwargs.get("user")
+
+    with transaction.atomic():
+        # Delete OAuth tokens associated with the user.
+        retire_dot_oauth2_models(user)
+
+        # Unlink LMS social auth accounts
+        UserSocialAuth.objects.filter(user_id=user.id).delete()
+
+        # Delete the user
```
jignaciopm: I think we should delete these kinds of comments

mariajgrimaldi: Done! Thanks

jignaciopm (dismissed):

jignaciopm (approved):

